PR #460, opened 2015-01-23
Make IP addresses into whois links

petecooper:
I chose IANA on the basis they're not tied to a particular registrar and seemed to be as neutral or independent as they come.

### DIFF
--- a/textpattern/include/txp_discuss.php
+++ b/textpattern/include/txp_discuss.php
@@ -410,7 +410,7 @@ function discuss_list($message = '')
                     txpspecialchars(soft_wrap($web, 15)), '', 'txp-list-col-website discuss_detail'
                 ).
                 td(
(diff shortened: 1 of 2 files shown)


rwetzlmayr:
I've tried to query IANA for a [well-known IP address](https://www.iana.org/whois?q=37.153.96.164) and all I got was "Ask RIPE" ;)
Would you be able find a slightly less US-centric whois service?

petecooper:
Yes, that's fair. I'll switch it to another provider.

petecooper:
OK, switched to DomainTools in the absence of a more obvious solution.

Wordpress uses ARIN for similar functionality, but that's also US-centric and provides the same level of info as IANA, broadly. DomainTools is what I've been using for years, and never had a problem with it.

PR #474, opened 2015-04-29
Tidy up `CONTRIBUTING.textile`

petecooper:
This should really be a `.txt` file instead of a `.textile`.

Either that, or the `README` should be Textile'd, too.


philwareham:
Thanks as usual, Pete!

PR #477, opened 2015-04-29
Tidy up on `README.textile`

petecooper:


philwareham:
Thanks Pete, keep going!

Bloke:
What was the eventual verdict on a [symlink for the latest version](http://forum.textpattern.com/viewtopic.php?id=41629)? Is it worth it? I suppose it just pushes the onus for keeping the symlink updated on the release / build process instead of the documentation. But at least the build process can be automated more easily than remembering to alter the docs by hand in all places the latest version is mentioned.

petecooper:
I reiterate my support for a symlink. Presumably there's shell access to the web server? If so, a symlink is a trivial thing to add to the process.

If GitHub is updated as part of the build process, then presumably `https://github.com/textpattern/textpattern/releases/latest` could be substituted.

Bloke:
I'd like to see this too. Someone (Robert? Phil? Jukka?) have server access, I don't.

philwareham:
Both Robert and I have server access - although I would prefer that Robert tackles this, as it's out of my comfort zone.

petecooper:
We're veering off topic, so I created https://github.com/textpattern/textpattern/issues/478 to gather this together.

PR #482, opened 2015-05-07
Have setup use minified CSS.

petecooper:
Min'd version is about 30% smaller.


philwareham:
Ah yes, that was missed when I minified. Thanks!

PR #483, opened 2015-05-07
Parser optimisations

vanmelick:
This patchset contains the parser optimisations that were tested in the following forum topic.:
http://forum.textpattern.com/viewtopic.php?id=43518

PS. this does not include the prefix handling changes that were proposed at the beginning of that topic. I'll issue a separate pull request for those.


vanmelick:
Please ignore this one.

PR #484, opened 2015-05-07
Parser optimisations

vanmelick:
This patchset contains the parser optimisations that were tested in the following forum topic.:
http://forum.textpattern.com/viewtopic.php?id=43518

PS. this does not include the prefix handling changes that were proposed at the beginning of that topic. I'll issue a separate pull request for those.


bloatware:
I thought namely of people publishing txp tips like '&lt;txp:tag /&gt;'. There is no need to parse if there is no real txp tags in the articles body.

philwareham:
@rwetzlmayr or @Bloke - do you want to check this patch over for a merge? I can't check it myself as it's not my area of understanding. Cheers!

rwetzlmayr:
@vanmelick Is this pull request ready for merging or will you add more commits here?

textpattern/vendors/Textpattern/Tag/Registry.php
```diff
@@ -74,11 +74,9 @@ public function register($callback, $tag = null)
      * @return string|null The tag's results
      */
 
-    public function process($tag, array $atts = null, $thing = null)
+    public function process($tag, array $atts = array(), $thing = null)
     {
-        if ($this->isRegistered($tag)) {
```
rwetzlmayr: Where did the tag registry go?

vanmelick: As I said in commit 35d62a4:
Don't check ifRegistered twice. It's already done in txplib_publish.php before calling this process() function. I was tempted to leave it in here and check if the return value was 'null' in txplib_publish.php, but some taghandlers (txp:variable!) don't return a value (=null) so there wouldn't be any way to differentiate it from a non-registered tag.

vanmelick:
I wouldn't merge it yet. Perhaps best to first decide which parts you want. I can create a new, clean branch if necessary or perhaps it is possible to cherry pick commits:
* Minor speed optimisations (I'd say yes)
* Parser rewrite and parse_else and change the default behaviour of EvalElse so that it simply calls parse_else (not so sure about this. Yes. it tends to be faster on complex pages, especially when looping a lot, but enough to justify such a change?)
* Short tags (<abc::tag> instead of <txp:abc_tag>).
* Splat optimisations (shortens runtime in some cases, not always... not yet committed.)

vanmelick:
@rwetzlmayr, see above comments and also http://forum.textpattern.com/viewtopic.php?pid=290694#p290694

textpattern/vendors/Textpattern/Tag/Registry.php
gocom: If you want to remove the second test, then you should throw an exception in ```Textpattern_Tag_Regisry::process()``` and do the check in it.

The function shouldn't generate unexpected, non-catchable errors when an invalid tag name is passed to it. Instead, use exceptions if you want to eliminate the second check, please.

philwareham:
Bumping this?

vanmelick:
How about replacing `class Textpattern_Tag_Registry` with

``` php
class Textpattern_Tag {
  public function register
  public function parse
  public function parseElse
}
```
By putting the tag registry and parser in a single class, I think we can optimize it even more. Yay or nay?

Bloke:
Hey that sounds like a plan! Presumably we'd just retain the existing `splat()` and `parse()` etc functions as stubs for backwards compatibility, and gradually move the core over to use the new class methods. I like the sound of this.

As an aside (perhaps this could be a separate Issue, but I'm not sure if it might fall under this one), what sort of effort is required to realise something along the lines of the [namespaced attributes](http://forum.textpattern.com/viewtopic.php?pid=268851#p268851) to which Jukka refers? Having a set of extensible global attributes that perform common actions (like trim) baked into the parser potentially saves attribute bloat in future (and plugins would get the facility for free). Is it worth considering, either here or as a separate issue?

vanmelick:
I think that can be added. We could then also not just register tags, but also global tag attributes, both built-in and supplied by plugins. We'd have to decide if we only want to accept them in the form `txp:attribute="value"` or also `txp:attribute' (like: `txp:trim`).

Bloke:
If it doesn't slow it down too much, both `txp:attribute` and `txp:attribute="value"` might be handy, otherwise we get into the situation of having to do `txp:trim="1"` which feels very HTML4ish. But I'd rather keep the speed up for the sake of typing 4 extra characters.

If it's possible to process the attributes in the order they appear in the tag, it gives a nice level of control to designers to influence what happens and when. But again, speed wins here, so whichever works best is fine.

`wraptag` and `class` attributes are tricky as you say. If we end up processing them in order, what happens if `class` appears before `wraptag`? Arguably, it means to assign the class to the tag's markup and then apply the classless wraptag as a container. If the administrator switches attribute order so that `wraptag` comes first, the `class` would be applied to that. If attribute order is not taken into account, it effectively 'makes a note' of the class and applies it to the wraptag when that attribute is encountered.

Both methods have drawbacks and advantages. Attribute order breaks the HTMLesque vibe, since HTML itself has no notion of attribute order. For consistency with that, order should be unimportant, but then Txp is a step up in functionality from plain HTML so maybe it's acceptable? I dunno...

I'm comfortable with whichever mechanism works reliably and quickly without being too confusing. If that means we limit the scope of global attributes to the 'on/off' variety and leave regular attributes for the complex stuff, so be it.

Maybe trying a few flavours of attribute style to see which feels better will help?

philwareham:
This pull request probably needs to be broken up into separate patches, each of a smaller scope.

philwareham:
@Bloke @rwetzlmayr @vanmelick just trying to mop up the various pull requests - is anything from this PR pertinent to the latest code? If so, can it be refactored to fix the conflicts and merged?

bloatware:
I would love to see it in 4.6. I have compared the latest Ruud's version against the current master. On a 100 articles listing with 15 txp tags in the article form it gives ~50% global runtime acceleration (0.75s -> 0.52s). But it is about 100 commits behind master, so this ratio may be biased. Any chance of update, please?

Other thing: I think the old `EvalElse` should be kept in some form for a moment. Textpattern does not need it anymore, but some wide-spread plugins (notably `smd_query`) do. Even if they were quickly patched, we should leave users few months to update them.

bloatware:
The difference is less noticeable (~15%) in live mode. But what really harasses me is that 4.5.7 seems to be about 40% faster, whatever setting (empty page or heavy article listing). Could someone confirm?

bloatware:
Got it: `call_user_func($func, $atts)` (used in `Registry`) is much slower than `$func($atts)` in some setups. Do we really need it?

vanmelick:
call_user_function also works when calling object methods instead of just calling plain functions. I suspect this extra functionality comes at a price.

bloatware:
Yes, but do we (plan to) use them? Then we could at least treat plain functions differently.

vanmelick:
Checking if it's a plain function call also takes time. According to this benchmark, the difference between a variable function call and call_user_func not that big:
https://gist.github.com/stanlemon/9847488
Yes, it takes 50% longer, but that's just when calling a completely empty function.

bloatware:
It is server/version/etc dependent. On my laptop I pass from 575ms to 325ms to process 15 tags article form 100 times. Checking is done only when registering, it's just a flag afterwards.

vanmelick:
What exactly is the difference in code between those two numbers, because it can't be variable function vs call_user_func. All benchmarks I've seen show just 1 or 2 microseconds difference. 15 tags times 100 times 2 would be a 3 ms difference. Not 250ms.

bloatware:
It's just `return (string) $this->tags[$tag]((array)$atts, $thing);` vs `return call_user_func($this->tags[$tag], (array)$atts, $thing);` in `Registry->process`. And I just output 100 articles through 15 tags form, getting ~60% difference. I have no ftp access to a live server atm, will retest next week.

bloatware:
@vanmelick Still chasing, but

- you are right, there is no noticeable difference between variable function call and `call_user_func` on a typical lamp server;
- I am right too, the long-run difference is about 30% on my wamp laptop, very weird;
- since the last [merge](https://github.com/textpattern/textpattern/pull/484/commits/112fa7efdf9c39cb126652e5a4a06131ad1d3e24), the [parser](http://www.iut-fbleau.fr/projet/test/textpattern-parser/articles/) branch is 10% slower than before (both lamp and wamp). It still is 10% faster than [master](http://www.iut-fbleau.fr/projet/test/textpattern-master/articles/), but 10% slower than [4.5.7](http://www.iut-fbleau.fr/projet/test/textpattern-4.5.7/articles/). I have tried to comment out `$trace` calls (~3000), this does not seem to help.

PR #488, opened 2015-06-25
Drop Apache 1.x support

petecooper:
Background: http://forum.textpattern.com/viewtopic.php?id=45826


philwareham:
Since no code has changed, I don't think we need to explicitly state in the README that Apache 1 is not officially support (that would be best suited to mention at documentation repo).

petecooper:
Hmm, OK. Fair point. How about `System requirements` instead of `Changed`?

philwareham:
But nothing actually happened, no code has changed that prevents Apache 1.x from running - or am I missing a previous commit that negates use of Apache 1.x?

petecooper:
You are correct, you're not missing anything - no code change. It's a documented _support_ change.

petecooper:
Yeah, you're right - I'll close.

PR #494, opened 2015-07-08
Debug log: trace MemUsage

makss:
This patch allows you to trace memory usage and help you find a Form/Tag/Plugin/SQL query that takes a lot of memory.

Forum thread
http://forum.textpattern.com/viewtopic.php?pid=292775#p292775


philwareham:
@rwetzlmayr @Bloke Do you want to verify this one? It's out of my knowledge area. Cheers!

Issue #497, opened 2015-07-10
More detailed trace log


makss:
I propose to include in the trace log all important events and challenges. I think that the admin interface also need the trace log.

Possible improvements:
* Admin interface trace log  - (done, sent Pull request)
* Trace callback_event  - (It's fantastic, especially in the admin interface. I really liked it. It will be next Pull request)

In what other components you want to add trace log?

PR #498, opened 2015-07-10
Introduce Admin Side trace log

makss:
Issue #497 

Slightly rewritten code.


Bloke:
Wow, awesome. Having memory and query usage on the admin side is going to be a huge help when developing plugins. You rock.

Could you please prefix the constants with `TEXTPATTERN_` just for crude namespace control. Then I'll merge this in, thanks.

makss:
It seems all corrected, Is possible merge.

Issue #196, opened 2014-10-04
do_list_unique() - more correct SQL queries


ghost:
_From [zasl...@gmail.com](https://code.google.com/u/112308807792196922428/) on May 22, 2012 12:11:05_

What steps will reproduce the problem?

    <txp:article_custom section="ss,ss,ss,ss,ss" category="cc,cc,cc,c2,c2" author="aa,aa,aa,aa" id="1,2,3,4,3,3,3" keywords="kk,kk,kk,kk" />

Duplicate values ​​- is the output of other tags.

What is the expected output?

    select *, unix_timestamp(Posted) as uPosted, unix_timestamp(Expires) as uExpires, unix_timestamp(LastMod) as uLastMod from txp_textpattern as textpattern where 1=1 and Status = 4 and Posted <= now() and (now() <= Expires or Expires = '0000-00-00 00:00:00') and ID IN (1,2,3,4) and (Category1 IN ('cc','c2') or Category2 IN ('cc','c2')) and Section IN ('ss') and AuthorID IN ('aa') and (FIND_IN_SET('kk',Keywords)) order by field(id, 1,2,3,4) limit 0, 10

What do you see instead?

    select *, unix_timestamp(Posted) as uPosted, unix_timestamp(Expires) as uExpires, unix_timestamp(LastMod) as uLastMod from txp_textpattern as textpattern where 1=1 and Status >= 4 and Posted <= now() and (now() <= Expires or Expires = '0000-00-00 00:00:00') and ID IN (1,2,3,4,3,3,3) and (Category1 IN ('cc','cc','cc','c2','c2') or Category2 IN ('cc','cc','cc','c2','c2')) and Section IN ('ss','ss','ss','ss','ss') and AuthorID IN ('aa','aa','aa','aa') and (FIND_IN_SET('kk',Keywords) or FIND_IN_SET('kk',Keywords) or FIND_IN_SET('kk',Keywords) or FIND_IN_SET('kk',Keywords)) order by Posted desc limit 0, 10

Solution:

Replace the call to do_list() do_list_unique()

    function do_list_unique($list, $delim = ','){
        return array_unique(do_list($list, $delim));
    }

This was tested in aks_article plugin.

_Original issue: http://code.google.com/p/textpattern/issues/detail?id=192_

makss:
It's been three years, perhaps the time has come to apply this patch.  :-)

If the objection is not, then I will prepare and test the patch, as described in the first post.
1. Separate function do_list_unique()
[...]

PR #499, opened 2015-07-10
Introduce do_list_unique()

makss:
Issue #196

### DIFF
--- a/textpattern/lib/constants.php
+++ b/textpattern/lib/constants.php
@@ -783,14 +783,16 @@ define('TEXTPATTERN_LANG_OWNER_SYSTEM', '');
 define('TEXTPATTERN_LANG_OWNER_SITE', 'site');
 
 /**
- * Strip empty HTML attribute values.
+ * Strip empty values.
  *
  * @since   4.6.0
  * @package HTML
- * @see     join_atts()
+ * @see     join_atts(), do_list_unique()
  */
 
-define('TEXTPATTERN_ATTS_STRIP_EMPTY', 0x1);
+define('TEXTPATTERN_STRIP_NONE',         0);
+define('TEXTPATTERN_STRIP_EMPTY',        0x1);
+define('TEXTPATTERN_STRIP_EMPTY_STRING', 0x2);
 
 /**
  * Sends an adaptive announcement.
@@ -836,14 +838,3 @@ define('TEXTPATTERN_ANNOUNCE_ASYNC', 0x4);
  */
 
 define('TEXTPATTERN_ANNOUNCE_REGULAR', 0x8);
-
-/**
- * Strip array/string flags
- *
- * @since   4.6.0
- * @see     do_list_unique()
- */
-
-define('STRIP_NONE',         0);
-define('STRIP_EMPTY',        0x1);
-define('STRIP_EMPTY_STRING', 0x2);
--- a/textpattern/lib/constants.php
+++ b/textpattern/lib/constants.php
@@ -836,3 +836,14 @@ define('TEXTPATTERN_ANNOUNCE_ASYNC', 0x4);
  */
 
 define('TEXTPATTERN_ANNOUNCE_REGULAR', 0x8);
+
+/**
+ * Strip array/string flags
+ *
+ * @since   4.6.0
+ * @see     do_list_unique()
+ */
+
+define('STRIP_NONE',         0);
+define('STRIP_EMPTY',        0x1);
+define('STRIP_EMPTY_STRING', 0x2);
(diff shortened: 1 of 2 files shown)


Bloke:
Fantastic, thank you.

To avoid polluting the global scope with constants that may clash with other projects, would you please prefix the constants with `TEXTPATTERN_`.

Also, take note that we already have `TEXTPATTERN_ATTS_STRIP_EMPTY` with the same value `0x1` as your proposed new constant. Is there any mileage in renaming `TEXTPATTERN_ATTS_STRIP_EMPTY` as `TEXTPATTERN_STRIP_EMPTY` (and altering `join_atts()` accordingly, which I believe is the only place that uses the constant at the moment). Then you could introduce the other two constants alongside it.

Does that make sense, or is it a bad idea?

Oh, and would you please tweak the function a teensy bit for our PSR-2 house coding style, specifically with regards the [closure](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#6-closures). Thanks.

makss:
You're right, need to use a prefix `TEXTPATTERN_` and rename `TEXTPATTERN_ATTS_STRIP_EMPTY` to `TEXTPATTERN_STRIP_EMPTY`

Is it possible to somehow check their code for compliance with PSR-2? How is this best done?
I'm a newbie at GitHub.

Bloke:
I think Phil uses [PHP-CS](https://mattstauffer.co/blog/using-php-cs-fixer-to-fix-up-your-php-code) because there's a `phpcs.xml` file in the Textpattern root directory.

If you use Sublime Text there's a plugin for PHP-CS that checks syntax (very handy) and PSR compliance on file save (in theory, but I have it installed and its never complained even if I violate the spec, so it probably needs some configuration in my editor). It does have some PEAR dependencies you need to install first.

philwareham:
Actually @gocom set up the `phpcs.xml` file.

I do run [PHP-CS-Fixer](http://cs.sensiolabs.org) over the entire project occasionally and fix things that are obvious and that I understand. I don't have it set up to monitor my code as I write it though.

Bloke:
I'd like PHP-CS to auto-check (to some degree,as long as it doesn't slow things down when in the behemoth `txplib_misc.php`) but it doesn't appear to work. It does the syntax checking, just not the PSR-2 thing even though there's a config option set to tell it to do so. Meh. Must be something I missed.

makss:
Running `PHP-CS` with the parameters described in `CONTRIBUTING.textile` showed a bad result, continuous errors and warnings.

Running `PHP-CS-Fixer` with autocorrection gave an excellent result. Almost all of the code corresponds PSR-2 standard.
To run as follows:
```
$ php-cs-fixer fix textpattern --level=psr2
   1) rpc/TXP_RPCServer.php
   2) sites/site1/public/index.php
   3) sites/site1/admin/setup/index.php
   4) sites/site1/admin/index.php
   5) sites/site1/private/config-dist.php
   6) textpattern/lib/classTextile.php
   7) textpattern/lib/txplib_db.php
   8) textpattern/lib/class.thumb.php
   9) textpattern/lib/txplib_misc.php
  10) textpattern/lib/IXRClass.php
  11) textpattern/lib/PasswordHash.php
  12) textpattern/theme/classic/classic.php
  13) textpattern/theme/hive/hive.php
  14) textpattern/theme/remora/remora.php
  15) textpattern/update/_to_4.4.0.php
  16) textpattern/update/_update.php
  17) textpattern/update/_to_4.4.1.php
  18) textpattern/index.php
  19) textpattern/config.php
  20) textpattern/config-dist.php
  21) textpattern/include/import/import_mt.php
  22) textpattern/include/import/import_blogger.php
  23) textpattern/include/txp_tag.php
  24) index.php
  25) css.php
Fixed all files in 57.436 seconds, 33.500 MB memory used
```

It is basically a very small correction, then add a line break(spaces) or replace `NULL` for `null`. The code is almost always used `null`, but in several places use uppercase. There are a few comments on the `objects`, but in general, after some revisions will be to put the entire project on the AutoCorrect using `PHP-CS-Fixer`. Is possible not run it at each commit, and periodically for the entire project.

Show log autocorrections `PHP-CS-Fixer`, to assess the amount of work?
Autocorrection `PHP-CS-Fixer` I liked, they save time. :)

philwareham:
That's because I have already run it over the project a few times and fixed a lot of the errors it throws up. :)

The @public@ and @private@ functions and a few other things it automatically corrects I have refrained from doing (including making @NULL@ lowercase) as I don't fully understand what they are or what ramifications it would have by changing them

makss:
@Bloke , The patch seems corrected, check please.

PR #504, opened 2015-07-15
Option: Preload all forms in a single query

makss:
I propose to allow the user to include preloading the forms yourself. (Admin / Preferences / Publish / `enable_preload_form` Pref. RadioButton: Yes/No)
Preloading all forms will be when you first load any form. Backward compatibility is completely satisfied, because do not change the call parameters `fetch_form()`.

Even on the default site(after Setup) saves 4 SQL query per page.
Counted for the page http:// ... /articles/1/welcome-to-your-site

http://forum.textpattern.com/viewtopic.php?id=39374

### DIFF
--- a/textpattern/lib/txplib_misc.php
+++ b/textpattern/lib/txplib_misc.php
@@ -4221,7 +4221,15 @@ function EvalElse($thing, $condition)
 
 function fetch_form($name)
 {
+    global $prefs;
     static $forms = array();
+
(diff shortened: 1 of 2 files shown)


makss:
Probably you do not need to apply this patch.  I close this Pull request.

A more flexible option is to change the type of a variable `$forms` of `static` to `global`.
It is allowed to do `Preload all forms` using the plugin from sql or file system..

PR #505, opened 2015-07-19
Tweak pluggable_ui

bloatware:
Avoid multiple `$default` output if more than one plugin modify some ui element. Pass the output of each plugin as default to the next one.


rwetzlmayr:
Thanks, nice one. 

I think it would be cleaner to add an option to ````callback_event()``` to achieve the desired behaviour than to duplicate its code in pluggable_ui().  

Is this feasible?

bloatware:
Nice proposal, thanks, much cleaner indeed. Could we use `debug_backtrace()` in `callback_event()`, and modify `$argv` if it's called from `pluggable_ui()`? Looks feasible, if it's ok performance-wise, will test.

rwetzlmayr:
I'd rather not introduce another global or depend on contents of the call stack. This reduces testability and code readability.

Why wouldn't you use an additional optional parameter to ```callback_event()```. like ```callback_event($event, $step = '', $pre = 0, $flag_you_name_it = false)```?

bloatware:
I may miss something obvious, but how would it be re backward compatibility? The actual `callback_event()` would pass this `$flag` as third argument to plugins, but they expect other data (often `$rs` array). We could strip 4 arguments from `callback_event()`, but wouldn't this require rewriting about 56 `callback_event()` core calls, plus all plugins that use it? We could somehow integrate `$flag` into `$pre` too, but it looks weird. We could also check if `$event` ends with `_ui`, but this is fragile.

On the other hand, what this option should do in general? In the case of `pluggable_ui()` only one additional string argument (`$default`) is passed as `$argv` to plugins, so `callback_event()` can simply replace it with return values. But in most cases `$argv` is not a string, but plugins return values are generally strings. Have you some examples in mind where this option could be useful?

bloatware:
These latest commits are involuntary (I thought only my fork was concerned), but tested and working so far. The hack is to pack new flag(s) into `$pre` in a backward-compatible way.

bloatware:
And why this one lands here? I've got yet a lot to learn about Github...

PR #508, opened 2015-07-28
Avoid links to expired articles

makss:
Discussion: http://forum.textpattern.com/viewtopic.php?id=46062

* For expired articles is returned link to `site_url/#expired_article`
* Show warning `permlink_to_expired_article` if `production_status != 'live'` && `txpinterface == 'public'`

### DIFF
--- a/textpattern/lib/txplib_misc.php
+++ b/textpattern/lib/txplib_misc.php
@@ -5380,9 +5380,9 @@ function permlinkurl($article_array)
         return $permlinks[$thisid];
     }
 
-    if (empty($prefs['publish_expired_articles']) && !empty($expires) && $expires<time()) {
-        if ($production_status != 'live') {
-            trigger_error(gTxt('permlink_to_expired_article')." ID='{$thisid}'", E_USER_NOTICE);
+    if (empty($prefs['publish_expired_articles']) && !empty($expires) && $expires < time()) {
+        if ($production_status != 'live' && txpinterface == 'public') {
+            trigger_error(gTxt('permlink_to_expired_article', array('{id}' => $thisid)), E_USER_NOTICE);
         }
 
         return $permlinks[$thisid] = hu.'#expired_article';
--- a/textpattern/lib/txplib_misc.php
+++ b/textpattern/lib/txplib_misc.php
@@ -5332,7 +5332,7 @@ function permlinkurl_id($id)
 /**
  * Generates an article URL from the given data array.
  *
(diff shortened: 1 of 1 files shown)


Bloke:
Great stuff, thanks. I like the optimisations in https://github.com/makss/textpattern/commit/96ded5716d15a1a7f3886edcf9b804c528df17d0. Much cleaner.

Couple of things to note about https://github.com/makss/textpattern/commit/9c901bfd6e603df9b5b52b4c5f4d7270f4bdd265:

* This function is also used on the admin side where we probably don't want it to trigger. For example, set up a few articles with expiry times in the past, then on the Articles panel, do any bulk operation on a bunch of the articles, including the expired ones: you'll get the error triggered if your production status is not Live.
* When triggering a notice, it's better to include supplemental info as params to `gTxt()` so they get passed to the string for replacement. e.g.:

    trigger_error(gTxt('permlink_to_expired_article', array('{id}' => $thisid)), E_USER_NOTICE);

But on the whole I think this patch is a step in the right direction, so if you could fix these minor issues I'm happy to merge it in.

Oh, and btw, PSR-2 prefers spaces around operators such as `<` :-)

makss:
You are right, a warning is necessary only on Public side. Thank you for the hint of gTxt().

rwetzlmayr:
> For expired articles is returned link to site_url/#expired_article

I think we already had the proper solution regarding our reply to requests for expired articles. We sent a '410 Gone' response code in conformance with [RFC7231](https://tools.ietf.org/html/rfc7231#section-6.5.9).

Switching from a real URL to the fake fragment marker ``#expired_article`` in the URL will disturb monitoring and analytics tools and generally make Textpattern an worse web citizen.

makss:
Referring to expired articles - this is a mistake web developer. This patch provides a warning about such links and makes the lesser of all evils - refers to the root of the site with a marker `/#expired_article`. This marker will be visible in web analytics and webmaster will draw attention to a problem with the site.

It would be possible to add `id` article in marker, but I think it's unnecessary.

rwetzlmayr:
> This marker will be visible in web analytics

This is an assumption that only applies to analytics with client-side sensors (e.g. Google Analytics).

The fragment identifier is not sent to the server. It is therefore not visible in the server's access logs, the Textpattern log panel, and the like.

makss:
I will not object to the removal of the marker `#expired_article`, important to remain alert `trigger_error`.

Issue #398, opened 2014-10-04
Optimize getNextPrev() function


ghost:
_From [haut.b...@aliceadsl.fr](https://code.google.com/u/106452029543525188687/) on August 29, 2013 17:20:19_

If "section" field were included in the array output by getNextPrev() function, then link_to_prev/next() functions could use permlinkurl() instead of permlinkurl_id(), sparing a db query.

_Original issue: http://code.google.com/p/textpattern/issues/detail?id=394_

vanmelick:
Pull request sent.

rwetzlmayr:
Closed via #512.

PR #512, opened 2015-08-09
Fix issue #398: Optimize getNextPrev() function

vanmelick:


rwetzlmayr:
@vanmelick getNeighbour() returns an [empty string](https://github.com/textpattern/textpattern/blob/master/textpattern/lib/txplib_publish.php#L175) when no neighbour exists. Your PR assumed a boolean ``false`` in that case (cf [link_to_next()](https://github.com/textpattern/textpattern/blob/master/textpattern/publish/taghandlers.php#L1641), [next_title()](https://github.com/textpattern/textpattern/blob/master/textpattern/publish/taghandlers.php#L1719) and friends).

Would we rather fix that by using a boolean return type in ``getNeighbour()`` which seems more natural, or check its return type in caller functions with ``is_array()`` which looks kind of odd?

vanmelick:
I agree. We should fix that by using a boolean return type. I've had a look at other functions in txp_misc.php and returning array|false is quite common there. (One exception being load_lang_event, which returns either array or empty string... and perhaps more interesting, that function doesn't appear to be used anywhere)

Issue #29, opened 2014-10-04
<txp:comment_x_tags/> don't allow closing without trailing space.


ghost:
_From [j...@beardedbaby.net](https://code.google.com/u/101554656051604297040/) on August 04, 2009 17:32:42_

What steps will reproduce the problem? 1. Use any txp:comment tag like so: <txp:comment_name_input />. it works in 4.0.8 and 4.2.0-
rc1.
2. Now, use the same tag without a trailing space: <txp:comment_name_input/>. Textpattern 
throws the following warning, and fails to render the specified comment-form output: "Textpattern 
Warning: tag does not exist  on line xxx".

As far as I can tell, you can use *any* other Textpattern or TXP plugin tag without the trailing 
space-- so far, I've only noticed the comment form tags being affected by this.

This behavior appears in 4.0.8 and 4.2.0-rc1.

_Original issue: http://code.google.com/p/textpattern/issues/detail?id=25_

vanmelick:
Fixed by pull request #513

bloatware:
Should it be closed, or we expect 4.5.8?

PR #513, opened 2015-08-11
Fix issue #29

vanmelick:
Create taghandlers for all comment form tags and popup_comments. This does involve moving some attributes from the main comment_form tag to the underlying comment form tags.


petecooper:
Adding a clickable link to https://github.com/textpattern/textpattern/issues/29 for completeness.

rwetzlmayr:
> This does involve moving some attributes from the main comment_form tag to the underlying comment form tags.

I grok from the diff that these atts are  'isize', 'msgcols', 'msgrows', 'msgstyle', 'previewlabel', 'submitlabel', 'rememberlabel', and 'forgetlabel'.

For the sake of upgrading users, would you mind keeping those as a dummy in ``comments_form()`` and add a notice a la...
``
trigger_error(gTxt('deprecated_attribute', array('{name}' => $foobar)), E_USER_NOTICE);
``

vanmelick:
Correct. 

I could add such a trigger_error, but are you sure that sends the correct message? Those attributes would no longer have any effect in comments_form. It's one step beyond deprecating them. Deprecated to me means: it still works in this version, but you should change it.

Currently, the user gets this message:
> Textpattern Notice: Unknown tag attribute: msgrows. while parsing form comments_display on page archive

Adding dummy attributes would result in this message:
> Textpattern Notice: msgrows attribute is deprecated. while parsing form comments_display on page archive

I'd rather document this in the release notes, but if you still want the trigger_error added, let me know.

vanmelick:
I'd also like you input on the attributes for the comment_remember tag. Because the label tags have moved to the tags where they belong, it was no longer needed to name them like "rememberlabel" anymore ('label' is enough). The only exception is comment_remember, where there are two label attributes. Given the name of the tag, I renamed 'rememberlabel' to 'label' there, but left 'forgetlabel' as is. The alternative there would be to keep 'rememberlabel' as well (and no 'label' attribute).

rwetzlmayr:
> ...are you sure that sends the correct message?

I am not sure. I'd rather have a "real" b/c fix.

> Deprecated to me means: "it still works in this version, but you should change it."

This would be the preferable way to deal with it. With the current solution we make legacy users suffer from a fix to an internal deficiency which didn't (really) hurt them in the past. We are basically changing one annoyance in user support ("why can't comment-form tags have != 1 trailing spaces before the tag closes?") with another ("why is my comment form broken since I upgraded?").

I think we should bite the bullet and carry these deprecated attributes over from the enclosing ``comments_form`` tag to the interior form elements even if this means introducing another global, e.g. ``$thiscommentsform``? We could deprecate this clutch immediately and remove it again in a future release. It's complicated...

rwetzlmayr:
> I'd also like you input on the attributes for the comment_remember tag. [...] The only exception is comment_remember, where there are two label attributes. Given the name of the tag, I renamed 'rememberlabel' to 'label' there, but left 'forgetlabel' as is. 

-1 

> The alternative there would be to keep 'rememberlabel' as well (and no 'label' attribute).

+1

(gut feeling, no rationale)

vanmelick:
Both things should be fixed now in the commit I just did.

vanmelick:
I did take the opportunity to remove the style attribute from the comment_message_input tag. It still works (deprecated) if set in comments_form, but for future use, people should target the textarea directly via a stylesheet instead of via an inline style.

vanmelick:
To help compare:
`<comment_remember rememberlabel="Remember" forgetlabel="Forget" />` (as it is now... a lot of 'remember' in there.)
`<comment_remember label="Remember" forgetlabel="Forget" />`

philwareham:
Can someone open an issue at the documentation repo and list the tag changes this commit introduces. I will then amend the relevant tag documentation.

Thanks.

vanmelick:
If you tell me where that repository is, then I'll do that.

philwareham:
Sure it's here: https://github.com/textpattern/textpattern.github.io

Issue #250, opened 2014-10-04
Double urlencode?


ghost:
_From [dev.nnnn@gmail.com](https://code.google.com/u/dev.nnnn@gmail.com/) on August 20, 2012 16:18:49_

What steps will reproduce the problem? 1.Make a category using multi-byte (eg."報告")
2.Publish an article using the category.
3.View the article from top page, and the url of the category looks wierd. What is the expected output? What do you see instead? The encoded url seems to be encoded double, which makes the url very messy even using clean url. Using the category example "報告", the expected/insted output is as follows.
EXPECTED: http://example.com/textpattern/&#37;E3&#37;82&#37;AB&#37;E3&#37;83&#37;86&#37;E3&#37;82&#37;B4&#37;E3&#37;83&#37;AA/&#37;e5&#37;a0&#37;b1&#37;e5&#37;91&#37;8a INSTEAD: http://example.com/textpattern/&#37;E3&#37;82&#37;AB&#37;E3&#37;83&#37;86&#37;E3&#37;82&#37;B4&#37;E3&#37;83&#37;AA/&#37;25e5&#37;25a0&#37;25b1&#37;25e5&#37;2591&#37;258a/ Please use labels and text to provide additional information. Using the language Japanese, the clean-url shall be " http://example.com/textpattern/カテゴリ/報告 ", but the url becomes " http://example.com/textpattern/カテゴリ/&#37;e5&#37;a0&#37;b1&#37;e5&#37;91&#37;8a ".
Using urlencode, encoding "&#37;e5&#37;a0&#37;b1&#37;e5&#37;91&#37;8a" becomes "&#37;25e5&#37;25a0&#37;25b1&#37;25e5&#37;2591&#37;258a", which can be seen in the example above.
Since the link itself is working properly, it seems more like an extra encoding error when addind a new category.
The encoded string also can be observed by looking through the admin panel to edit the category while the category name will show the encoded string in the text-field (expected to be decoded. Maybe a different issue?). The category title seems to be decoded as expected (see attached screenshot).

**Attachment:** [category_edit_Textpattern_CMS.png](http://code.google.com/p/textpattern/issues/detail?id=246)

_Original issue: http://code.google.com/p/textpattern/issues/detail?id=246_

vanmelick:
[...]

Issue #249, opened 2014-10-04
Using multi-byte characters in the name of section


ghost:
_From [dev.nnnn@gmail.com](https://code.google.com/u/dev.nnnn@gmail.com/) on August 20, 2012 07:51:43_

What steps will reproduce the problem? 1.Login to admin page. Create or edit a section.
2.Use an multi-byte character in the section name
3.Save and try to edit by clicking the section name What is the expected output? What do you see instead? EXPECTED: The edit page for the section
INSTEAD: almost nothing except with a version info, user name, and logout button Please use labels and text to provide additional information. When I look in the source, I can see something is not going right.
I'll paste the whole source I get.
It only happens when I use a multi-byte character in the section-name, and it does not occur when I use the same characters in the section-title.

</div><div class="txp-footer"><a id="mothership" href=" http://textpattern.com/ " title="go_txp_com" rel="external"><img src="theme/classic/carver.png" width="40" height="40" alt="Textpattern" /></a>
<p>Textpattern CMS &#183; 4.5.0-beta</p>
<p id="moniker">user <span>****</span><br /><a href="index.php?logout=1">logout</a></p>
<script type="text/javascript">

</script>

</div></body>
</html>

_Original issue: http://code.google.com/p/textpattern/issues/detail?id=245_

vanmelick:
Stuff got lost while transferring from Google Code. Let's try that again:

---

What steps will reproduce the problem?
1.Login to admin page. Create or edit a section.
2.Use an multi-byte character in the section name
3.Save and try to edit by clicking the section name

What is the expected output? What do you see instead?
EXPECTED: The edit page for the section
INSTEAD: almost nothing except with a version info, user name, and logout button

Please use labels and text to provide additional information.
When I look in the source, I can see something is not going right.
I'll paste the whole source I get.
It only happens when I use a multi-byte character in the section-name, and it does not occur when I use the same characters in the section-title.

> `</div><!-- /txp-body --><div class="txp-footer"><a id="mothership" href="http://textpattern.com/" title="go_txp_com" rel="external"><img src="theme/classic/carver.png" width="40" height="40" alt="Textpattern" /></a>`
> `<p>Textpattern CMS &#183; 4.5.0-beta</p>`
> `<p id="moniker">user <span>****</span><br /><a href="index.php?logout=1">logout</a></p>`
> `<script type="text/javascript">`
> `<!--`
> `textpattern.textarray = null`
> `// -->`
> `</script>`
> 
[...]

PR #514, opened 2015-08-14
fix issue #250 and show correct error message.

vanmelick:
Fixes issue #250 and shows the invalid name in the error message if a category cannot be created (instead of not showing it at all). See commit messages and issue #250 for more detail.


vanmelick:
I believe this also fixes issue #249

Issue #457, opened 2015-01-02
sync article and article_custom attributes or generate warnings


vanmelick:
See this topic
http://forum.textpattern.com/viewtopic.php?id=44115&p=1

The user has excerpted="y" as an attribute in the txp:article tag, which isn't valid, but doesn't generate a warning, thus causing confusion. There should be a warning about using an unknown attribute or txp:article should handle the attribute as txp:article_custom does.

Bloke:
Yes, this is always a source of annoyance since `<txp:article>` is just a specialised version of `<txp:article_custom>` and uses most of the same code. As you know, the upshot is that you can put any attributes you like that are valid for article_custom into a regular article tag, and they'll mostly be silently ignored.

This could probably be changed with some `if $(is_custom)` stuff to detect which attributes have been supplied and throw appropriate warnings, just never really investigated how best to untangle the two.

vanmelick:
See pull request. 

Additionally, I wonder why TXP accepts 'y' as a valid value for that attribute, while TXP uses 0 or 1 for boolean attributes in all other places. Shouldn't that be corrected as well for consistency?

rwetzlmayr:
Refs #511

rwetzlmayr:
> Additionally, I wonder why TXP accepts 'y' as a valid value for that attribute, while TXP uses 0 or 1 for boolean attributes in all other places. Shouldn't that be corrected as well for consistency?

+1

Bloke:
Closed by 26db1247d5049dd28b87d95d1573afb41f28fc2a

PR #515, opened 2015-08-14
Treat excerpted attribute in article_custom as boolean.

vanmelick:
While using 'y' as a value works (and keeps working), TXP standardises on '0' and '1' for boolean attributes. This patch makes it more consistent with the rest of the code base.

### DIFF
--- a/textpattern/include/txp_category.php
+++ b/textpattern/include/txp_category.php
@@ -485,7 +485,7 @@ function cat_event_category_create($event)
     $name = strtolower(sanitizeForUrl($title));
 
     if (!$name) {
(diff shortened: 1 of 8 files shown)


vanmelick:
See the discussion at the bottom of issue #457 (which can be closed if this is merged).

This does additionally require a change in the Wiki:
http://www.textpattern.net/wiki/index.php?title=article_custom

For the "excerpted" attribute it should read:
 excerpted="boolean"
    Whether to restrict to articles with an excerpt (or not).
    Values: 0 (also include articles without excerpt) or 1 (restrict to articles with an excerpt)
    Default: 0.

Issue #411, opened 2014-10-04
Signal web server about failed login attempts with a HTTP header


ghost:
_From [jukka.m.svahn](https://code.google.com/u/jukka.m.svahn/) on March 10, 2014 12:19:22_

We should signal the server about failed login attempts with a HTTP header, either with Status or otherwise. Currently we have no real way of tracking failed logins; there isn't even a dedicated URL to check for requests.

This would allow people to track failed login attempts and prevent brute force attacks. Tools such as Fail2Ban can be set to track Apache's error logs.

The headers we could send are a custom one (X-Textpattern-AuthAttempt: LoginFailure), or status 401 or 403. Each would allow proper logging and greatly enhance login security.

_Original issue: http://code.google.com/p/textpattern/issues/detail?id=407_

Bloke:
Closed by f95dac5edc321c8b50e2ab4c80ca9c7c895643e5

PR #516, opened 2015-08-14
Fix issue #411

vanmelick:
Generate a HTTP 401 response if a user fails to log in or tries to log in with a bad cookie. Fixes issue #411


vanmelick:
I hardcoded the status message. Could've used gTxt, but that doesn't seem wise in a response header (you never know what gTxt might return)

PR #517, opened 2015-08-14
Fix 'internal error' javascript alert

vanmelick:
The alert appeared after a failed login due to an missing parameter in a call to the mysqli_get_server_info function.

### DIFF
--- a/textpattern/lib/txplib_misc.php
+++ b/textpattern/lib/txplib_misc.php
@@ -4055,6 +4055,8 @@ function change_user_group($user, $group)
 
 function txp_validate($user, $password, $log = true)
 {
+    global $DB;
+
     $safe_user = doSlash($user);
     $name = false;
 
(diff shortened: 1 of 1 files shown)


vanmelick:
What surprised me in this error is how uninformative it was and how it was shown. It's generated in line 1612 in txplib_misc.php: gTxt('internal_error'). The javascript alert was inserted at the very beginning of the HTML code, before the `<html>` tag. I had to add $errno, $errstr, $errfile, $errline in the code to track down what triggered it. Not sure if there's a reason why such errors are handled this way.

rwetzlmayr:
> Not sure if there's a reason why such errors are handled this way.

Not sure either, but I assume that we would show explicit error messages only to authorized users (cf. check for 'debug.verbose' privilege in line 1609). which is nothing we can accomplish  while logging someone in in ``txp_validate()``. Maybe we should consider ``$production_status`` here.

Issue #261, opened 2014-10-04
load_lang() - get all untranslated strings from 'en-gb'


ghost:
_From [zasl...@gmail.com](https://code.google.com/u/112308807792196922428/) on August 30, 2012 12:07:48_

Get all untranslated strings from 'en-gb' language for admin/public interface.

ps: function load_lang_event() also something loads from txp_lang, but the current code this function is not used anywhere.

Patch attached.

**Attachment:** [txplib_misc_load_lang_r4090.patch](http://code.google.com/p/textpattern/issues/detail?id=257)

_Original issue: http://code.google.com/p/textpattern/issues/detail?id=257_

Bloke:
Closed by 33ac470979dd9fd854e7a1eef72160aa1dbfbcaf

Issue #103, opened 2014-10-04
Empty txp_lang items cause load_lang_event to fail


ghost:
_From [r.wetzlmayr](https://code.google.com/u/r.wetzlmayr/) on January 17, 2011 08:32:11_

== Symptom ==

jsoo reports at http://forum.textpattern.com/viewtopic.php?id=35305 :

The problem is that an empty (no name) item in txp_lang causes load_lang_event() to think that the language is not installed, so it defaults to en-gb.

== Reason ==

Languages on rpc.textpattern.com erroneously contain rows with empty 'name' columns. These are mirrored to individual RPC client sites via the language update mechanism.

== Solution ==

Defend against invalid langueage entries at all steps involved: 

* At the RPC server: Perform cleanup and reject future invalid data
* At the RPC client: Reject invalid data
* At the app core: Defensive use of language data

_Original issue: http://code.google.com/p/textpattern/issues/detail?id=99_

Bloke:
Closed by 33ac470979dd9fd854e7a1eef72160aa1dbfbcaf

PR #518, opened 2015-08-15
Fix issue #261: Reduce load_lang memory usage. 

vanmelick:
This applies the parts of the patch in issue #261 that result in reduced memory usage. The remainder of the patch in that issue was rejected, so applying this one would close that issue.
It also removes the unneeded "empty" call.

### DIFF
--- a/textpattern/lib/txplib_misc.php
+++ b/textpattern/lib/txplib_misc.php
@@ -497,7 +497,7 @@ function load_lang($lang, $events = null)
     }
 
     foreach (array($lang, 'en-gb') as $lang_code) {
-        $rs = safe_rows('name, data', 'txp_lang', "lang='".doSlash($lang_code)."'".$where);
+        $rs = safe_rows_start('name, data', 'txp_lang', "lang='".doSlash($lang_code)."'".$where);
 
         if (!empty($rs)) {
             break;
@@ -507,10 +507,8 @@ function load_lang($lang, $events = null)
     $out = array();
 
     if (!empty($rs)) {
-        foreach ($rs as $a) {
-            if (!empty($a['name'])) {
-                $out[$a['name']] = $a['data'];
-            }
+        while($a = nextRow($rs)) {
+            $out[$a['name']] = $a['data'];
         }
     }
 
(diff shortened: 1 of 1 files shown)


vanmelick:
Seems the 'empty' call was needed after all (#103), so it's back in there. 

What surprises me is that the optimization proposed in this pull request used to be part of TXP in the past, but somehow got removed along the way. Any reason for using safe_rows instead of safe_rows_start+nextRow here? In other places (like get_prefs) this isn't done (optimized safe_rows_start+nextRow) is used there.

makss:
Some convoluted code in function `load_lang()`.
Probably it is necessary replace
```
function load_lang($lang, $events = null)
{
    if ($events === null && txpinterface != 'admin') {
        $events = array('public', 'common');
    }
    $where = '';
    if ($events) {
        $where .= ' and event in('.join(',', quote_list((array) $events)).')';
    }
    foreach (array($lang, 'en-gb') as $lang_code) {
        $rs = safe_rows_start('name, data', 'txp_lang', "lang='".doSlash($lang_code)."'".$where);
        if (!empty($rs)) {
            break;
        }
    }
    $out = array();
    if (!empty($rs)) {
        while($a = nextRow($rs)) {
            if (!empty($a['name'])) {
                $out[$a['name']] = $a['data'];
            }
        }
    }
    return $out;
}
```

To:
```
function load_lang($lang, $events = null)
{
    if ($events === null && txpinterface != 'admin') {
        $events = array('public', 'common');
    }
    $out = array();
    $where = '';
    if ($events) {
        $where .= ' and event in('.join(',', quote_list((array) $events)).')';
    }
    foreach (array($lang, 'en-gb') as $lang_code) {
        $rs = safe_rows_start('name, data', 'txp_lang', "lang='".doSlash($lang_code)."'".$where);
        if (!empty($rs)) {
            while($a = nextRow($rs)) {
                if (!empty($a['name'])) {
                    $out[$a['name']] = $a['data'];
                }
            }
            return $out;
        }
    }
}
```

What better use `safe_rows` or `safe_rows_start` - I do not know.

makss:
BTW, what to check `if (!empty($a['name']))`  ?  It may be easier to do this in sql query?

```
$rs = safe_rows_start('name, data', 'txp_lang', "lang='".doSlash($lang_code)."'".$where." and name<>''");
```

vanmelick:
True. That does look better. Patch sent.

Issue #269, opened 2014-10-04
Suppressing errors if textpattern folder is corrupt


ghost:
_From [zasl...@gmail.com](https://code.google.com/u/112308807792196922428/) on September 06, 2012 08:57:35_

What steps will reproduce the problem? 1. Delete folder /textpattern
2. Open site via web

What do you see?
Many errors with real_full_path to textpattern: /var/www/txp_svn/
---
Warning: include(/var/www/txp_svn/textpattern/lib/constants.php) [function.include]: failed to open stream: No such file or directory in /var/www/txp_svn/index.php on line 68

Warning: include() [function.include]: Failed opening '/var/www/txp_svn/textpattern/lib/constants.php' for inclusion (include_path='.:/usr/share/pear:/usr/share/php') in /var/www/txp_svn/index.php on line 68

Warning: include(/var/www/txp_svn/textpattern/lib/txplib_misc.php) [function.include]: failed to open stream: No such file or directory in /var/www/txp_svn/index.php on line 69

Warning: include() [function.include]: Failed opening '/var/www/txp_svn/textpattern/lib/txplib_misc.php' for inclusion (include_path='.:/usr/share/pear:/usr/share/php') in /var/www/txp_svn/index.php on line 69

Fatal error: Call to undefined function txp_status_header() in /var/www/txp_svn/index.php on line 72
---

After patch will be displayed simple string:
---
config.php is missing or corrupt. To install Textpattern, visit textpattern/setup/
---

**Attachment:** [index_noerror_r4096.patch](http://code.google.com/p/textpattern/issues/detail?id=265)

_Original issue: http://code.google.com/p/textpattern/issues/detail?id=265_

rwetzlmayr:
Won't fix, see https://github.com/textpattern/textpattern/pull/521#issuecomment-132162081 for rationale.

PR #521, opened 2015-08-18
Fix issue #269

makss:
Fix issue #269

If the function `txp_status_header` is not available then returns the error status 200 instead of 503.
So was replaced `txp_status_header()` on` header()`.

### DIFF
--- a/index.php
+++ b/index.php
@@ -83,7 +83,7 @@ if (!isset($txpcfg['table_prefix'])) {
 }
 
 if (!isset($txpcfg['table_prefix'])) {
-    header("HTTP/1.0 503 Service Unavailable");
+    header('HTTP/1.0 503 Service Unavailable');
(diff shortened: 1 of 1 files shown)


rwetzlmayr:
As stated here https://github.com/textpattern/textpattern/issues/269#issuecomment-57904538, I still do not think that we can add really useful code to cater for people who, for whatever reason, voluntarily delete parts it for reasons beyond my imagination.

What is your real-life use case?

makss:
I always use a folder other than `/textpattern/` is why I saw this error.
The error message can not clean up, but it is desirable for the search engine bots to return the status of 503 instead of 200.

rwetzlmayr:
Thanks, but we do not try to fix bugs which stem from non-standard deployment practises, use of forks or modified versions of Textpattern, and do not affect the general audience. 

I'd suggest that you use your own fork, and enjoy the beauty of open source and libre licenses.

PR #522, opened 2015-08-18
Allow prefetch_forms via plugin

makss:
Change type variable `$forms` from `static` to `global`.
This will make using plugin preload forms of sql / file system / other source.

Discussion: http://forum.textpattern.com/viewtopic.php?id=39374

ps: Perhaps it will be useful for `themes branche` (preload theme forms).

### DIFF
--- a/textpattern/lib/txplib_misc.php
+++ b/textpattern/lib/txplib_misc.php
@@ -4192,7 +4192,7 @@ function EvalElse($thing, $condition)
 
 function fetch_form($name)
 {
-    static $forms = array();
+    global $forms;
 
     $name = (string) $name;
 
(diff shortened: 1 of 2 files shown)


rwetzlmayr:
Not convinced that this is the most elegant solution: http://forum.textpattern.com/viewtopic.php?pid=294182#p294182

Bloke:
I concur that this is not a desirable patch. As I said in my post on that thread Robert mentioned, polluting the global namespace further is something I would prefer to avoid. There has to be a better way.

makss:
@rwetzlmayr, @Bloke - You're right, let's try to find a better solution.

PR #526, opened 2015-08-19
Replace now() calls in SQL queries to allow query caching

vanmelick:
On websites that have a large number of articles, this dramatically reduces the time spent in SQL queries, because they can be cached. I've seen reductions from 90ms to 4ms, reducing total runtime by 75%.

This drops time accuracy to 30 seconds (+- 30sec) by default, which can be overridden by a hidden pref time_granularity. The default should be good enough for busy websites where pages get multiple hits per minute.


Bloke:
Sweet!

I made a half-hearted start on this in the [custom fields branch](https://github.com/textpattern/textpattern/commit/ab8c10ba47c8214cc68e138b8f6b9aca2d158c16) but it wasn't specifically targeted at all queries, just a new global (yuk) that meant time was "unified" for all CF-based queries and was therefore independent of page execution time.

Since Custom Fields nominally have posted and expiry times (at present, but it causes headaches so I might drop it!) reliance on wandering time could cause UI glitches.

Although there's the possibility that the UI may be drawn 'across' a time boundary, this pull request is far superior in many ways so I'll need to find a way to fold it into the custom fields branch in place of my hack job. Thanks.

makss:
Thanks for the patch.

I think that the default time can be safely increased to one hour. Increased time will allow more efficient use of query_cache. I do not think that can be critical exact time the article was published or hide.

vanmelick:
@Bloke: Yes. Pull request updated

@makss: I suspect those who publish stuff that is under NDA until a specific time might disagree. Increasing the default to 1 hour would result (worst case) in the article being visible half an hour too early (or too late). There may be other similar cases where this matters. Other than that, I'd agree with increasing the default.

vanmelick:
One could argue that people under NDA are a very very small minority of users. 
I was a bit worried about what would happen when a user publishes an article and then visits the website to see the result, thinking it might not show up due to caching, but I overlooked the fact that the cache is invalidated by table updates.

Stef, do you agree with updating the default to 1 hour?

Bloke:
Cache busting isn't my strong point, but given the cache is invalidated by table updates, an hour seems like a sensible default to me.

rwetzlmayr:
If I understand this correctly, with ``time_granularity`` set to 3600, a user might experience something along this lines:

1. User sets an article's "publish date" to "20:31".
1. User hits "Publish".
1. Due to rounding effects user does not see her article on the site until 21:00.

Is this correct? 

I do not understand what the fact "that the cache is invalidated by table updates" would change in this mathematical expression:

``
$now = (round(time()) / $granularity) * $granularity;
``

rwetzlmayr:
Opinion: As long as we provide a user interface which allows users to specify publish and expiry times by the second we must not have short-cuts in core which miss these time stamps by 30 minutes (avg).

This violates the principle of least surprise.

vanmelick:
Ah... default template in TXP doesn't use time="past", which is why I didn't see this happening during tests. 

Pull request modified. It now uses ceil() instead of round(), so you will always see an article you just posted with the current publication date instead of sometimes having to wait. It is switched off by default now, but it's a nice knob for tuning performance for those sites that need it.

rwetzlmayr:
What happens if the web server and the database server live in different time zones?

(I have at least one client project with such a setup)

vanmelick:
Without actually testing it...

Before: NOW() executes on the SQL server.
After: FROM_UNIXTIME(unixtime) executes on the SQL server. And unixtime is the same everywhere, regardless of timezone.

rwetzlmayr:
> Pull request modified. It now uses ceil() instead of round(), so you will always see an article you just posted with the current publication date instead of sometimes having to wait. 

That looks better. Let me pick a few nits, though, to help me understand the possible support issues we might face when this merge hits the public...

I ran this experiment:

- ``time_granularity`` is set to 3600 (I chose a big number to see the effects.)
- Wall clock / site clock says: 13:01
- I set an articles "Timestamp" field to: 13:15 (I want to publish in the future for whatever reason. We brag about this feature.)
- The article is visible on the site at the very moment i hit "Publish". In general, publishing occurs 0..30 minutes too early. Surprise!
- On the public site, it says "Posted a few seconds ago" until the wall clock passes the time from "Timestamp".
- I suspect that there might be similar surprises with the expiry timestamp.

> Even so, this is switched off by default now, but it's a nice knob for tuning performance for those sites that need it.

Our UI supports a granularity of 1 second. Wouldn't it be beneficial if `time_granularity` defaulted to 1, too? I grok that this wouldn't cause any side effects, eventually use cacheable queries w/ a query cache shared across multiple parallel requests, has very little computational overhead.

@vanmelick, what do you think?

vanmelick:
The 'posted a few seconds ago' string also occurs if you set time='any' on an article and use the 'since' dateformat (bug or feature?). In this case it's a nice side effect.

The "?" popup-help for this option should contain something like this, so the user is informed about the possible adverse affect of this setting:

> The time granularity affects how accurately Textpattern adheres to the publish / expiration times you specify. For example: setting it to 3600 (1 hour) will cause articles to appear or expire up to 1 hour too early. 

> However, setting a larger time granularity allow for more efficient SQL query caching on busy websites, making your website faster and reducing server load, which will certainly benefit busy websites.

> The default setting (1 second) is aimed at low traffic websites which require very accurate timekeeping. If your website gets a lot of visitors, we recommend increasing this setting to somewhere between 60 (1 minute) and 3600 (1 hour), depending on much you rely on having exact publish / expire times

A time granularity of 1 does have the benefit of now() generating the exact same value during that page request. The current (old) TXP versions could have multiple calls to now() resulting in differing values if you load a page right on the border of a switch in seconds. Patch committed.

makss:
Query_cache provides accelerated only **after the second request**. If `time_granularity` = 1 second, the need to speed up to get to in a second came to us two or more identical sql queries.

Yes, `ab` test will show us good acceleration, but on a real hosting it would not make sense. You can view the apache log and search request the same page in one second.

The bad news for `time_granularity` = 1: on hosting mysql query_cache one on all sites, intensively placing the cache our inquiries, we thus pushes saved queries to other sites. And because of this possible loss of performance on other sites.

Therefore, by default `time_granularity` best to disable or to put at least 60 seconds.
( Although I feel sorry for torturing mysql server so often ;-)  )

vanmelick:
Excellent point. Hadn't thought of that. So defaulting to 0 does make the most sense. I've modified my proposal for the popup help text in the previous comment

Bloke:
Curveball: what would be the effect of changing our Posted/Expiry widgets from the current set of six fields to either:

* a date field + time field, or
* a combined date/time field

and enforcing a `step` value of `time_granularity`? Would that at least give a visual indicator that your required timestamp is a guide only and if you need anything tighter, to reduce the pref value? Or is this whole cache issue just swapping one problem for another?

philwareham:
Regarding date/time widgets. Don't use the HTML5 date and time inputs, they are a mess right now. I have tried that in my admin update mockups and they simply don't work as you'd want them to.

I'm using patterns instead, for example:

    <input class="txp-form-field-input input-hour" id="hour" name="hour" type="text" size="2" maxlength="2" inputmode="numeric" pattern="(0[0-9]|1[0-9]|2[0-3])" title="Hour" aria-label="Hour" placeholder="hh" value="15">

Bloke:
Yes, I've pretty much come to the same conclusion, having been involved in implementing various date-based systems of late. I've defaulted to exactly the idea you mention, Phil.

Variance in browser support for the date/time types and non-uniform advertisement of a browsers' ability to handle them have rendered a potentially useful set of input widgets into a bit of a farce.

vanmelick:
@Bloke, in that case you'd have to change Posted/Expired dates in the database after modifying the time granularity.

Bloke:
True dat. I'll get back in my box.

bloatware:
Sorry if I miss something, but wouldn't it be better to store the "next upcoming" publication date somewhere (prefs?), and a) use it instead of NOW() if it's in the future; b) otherwise, retrieve a new "next upcoming" and go to a) ? One visitor will be penalized with an additional db query b), but it's not a big deal, and we gain in accuracy/efficacy.

vanmelick:
Interesting. 
* It's not just Posted, but also Expiration...  and the Publish date for files.
* It requires 4 DB queries (2 if you can retrieve the next posted/expired date in 1 query).
* now() would then not be a generic solution but closely tied to just the articles.

Pseudo code:

``` php
$old_now = fetch_pref();
if (time() > $old_now) {
  $new_now = min(fetch_next_posted(), fetch_next_expired(), fetch_next_file()) - 1; // unixtime.
  store_pref($new_now);
  return $new_now;
} else {
  return $old_now;
}

If you get multiple hits when time() = $old_now + 1, then each of those hits will keep updating the pref.

What would you store in the pref if there wasn't a future article nor an article that could expire?

**Comment by bloatware** [comment]
Point by point, without much thinking:

* `now($type)`, where `$type` could be article or file or...
* retrieving two fields in one query if both are outdated is feasible, though two separate queries could be faster, it needs testing.
* For me, ideally, one day everything should be articles, extended with custom fields.
* I don't see why, updating `$old_now` should take much less than 1s? We could also add few granularity seconds to `$new_now`.
* Some future dooms day?

**Comment by vanmelick** [comment]
the 'easiest' way would be to have three prefs, one for each column where now() is used (now($type), as you suggest).

If there's no future expiration/publish date you could set it to the year 2038, but you'd have to update the prefs each time you save an article or file on the admin side, by calling now($type, true), where type is `posted`, `expires` or `created` (if it's a file).

You'd get something like this:

``` php
function now($type, $update = false) {
  $old = get_pref('sql_now_'.$type, 0);
  $now = time();

  if ($now > $old or $update) {
    $now += 2; // time travel
    $table = ($type === 'created') ? 'txp_file' : 'article';
    $new = safe_field('unixtimestamp('$type') as utime', $table, $utime.' > '.$now.' order by '.$type.' asc limit 1');
    $new = (false === $new) ? 2147483647 : intval($new) - 1; 
    update_pref('sql_now_'.$type, $new);
    return $new;
  } else {
    return $old;
  }
}
```

Opinions, comments?

vanmelick:
Look in the comments on github for updated code proposal (you got an email with lots of bugs in the code).

bloatware:
Looks fine for me, needs testing now. You are right, we should update the prefs on *each* article/file post/save/delete. I can't see any negative drawback, just don't know how it copes with Stef's cf plans?

Bloke:
Ha! I've no idea how it fits with the CF plans either, so don't worry about it.

As far as I could see, to render the custom fields UI effectively and unambiguously, Txp needed a single, unified timestamp at the start of every page load. Otherwise, any delays in processing, network or rendering meant inconsistencies or errors. To do that I introduced a global in `index.php` upon which all time decisions were made.

This had the knock-on effect of me factoring out the use of `now()` in a bunch of database calls and injecting the variable `$txpnow` instead. At the time it had nothing to do with cacheability. Still doesn't, as I'm not very well versed in it all. But whatever path this patch takes, I'm sure we can figure something out for CFs.

makss:
> we should update the prefs on each article/file post/save/delete

Or just reset prefs (set to zero). The first call to the function `now()` prefs themselves updated.

makss:
A tiny remark. This sql query we will be executed only once, ie, cache does not need to.
```
    $where = 'utime > '.$now.' order by utime asc limit 1';
    $new = safe_field('unix_timestamp('.$type.') as utime', $table, $where);
```

Therefore, we can use the sql function `now()` which will eliminate caching.
```
    $where = 'utime > now() order by utime asc limit 1';
    $new = safe_field("unix_timestamp($type) as utime", $table, $where);
```

vanmelick:
I'm using $now there to circumvent possible differences in time between the database server and the server where the website is hosted. Pull request updated. This doesn't include caching queries for file lists. I'll do that separately to address another issue, once this request is accepted.

bloatware:
Nice request, I would just pass something like `$now < $posted < $old` (pseudo code) as `$update`, to spare few admin-side queries, but that's marginal. And an extra bonus of this request is the possibility to output forthcoming articles, setting the `$pref['sql_now_posted']` to some future date.

makss:
@vanmelick, there is a small typo, file  `_to_4.6.0.php`

```
// Add prefs to allow query caching when now() is used.
if (!get_pref('sql_now_publish')) {
    set_pref('sql_now_publish', time(), 'publish', PREF_HIDDEN);
```

Need `sql_now_posted`

vanmelick:
Thanks. Fixed.

vanmelick:
This set of commits now also fixes issue #531 (that part was really only one line that needed to be added)

vanmelick:
I'll work on fixing the merge conflicts

philwareham:
I'm going to merge this now as there hasn't been any further concerns raised.

Issue #528, opened 2015-08-20
URL to article 1 in feeds is http://example.com/textpattern/#expired_article


rwetzlmayr:
On a newly setup site, the feed item for the "Welcome to your site" article points to ``http://example.com/textpattern/#expired_article``.

The article in question is **not expired**, thus the proper URL would be ``http://example.com/articles/1/welcome-to-your-site``

rwetzlmayr:
Thanks, this was already fixed in https://github.com/textpattern/textpattern/commit/82393f068ea840ef1e4146e74efe80ca80881c23

rwetzlmayr:
Fixed in https://github.com/textpattern/textpattern/commit/82393f068ea840ef1e4146e74efe80ca80881c23

PR #529, opened 2015-08-20
Fix issue #528

makss:
Fix issue #528


rwetzlmayr:
Thanks, this was already fixed in https://github.com/textpattern/textpattern/commit/82393f068ea840ef1e4146e74efe80ca80881c23

PR #530, opened 2015-08-20
Escape double dash sequences inside comments...

vanmelick:
...because they signify an interruption in the comment text, which otherwise could be visible to people visiting the website

### DIFF
--- a/textpattern/lib/txplib_html.php
+++ b/textpattern/lib/txplib_html.php
@@ -1076,7 +1076,6 @@ function htmlPre($item, $atts = '')
 
 function comment($item)
 {
-    // Note that the replacement string is two &#8208 characters
     return '<!-- '.str_replace('--', '- - ', $item).' -->';
 }
 
--- a/textpattern/lib/txplib_html.php
+++ b/textpattern/lib/txplib_html.php
@@ -1077,7 +1077,7 @@ function htmlPre($item, $atts = '')
 function comment($item)
 {
     // Note that the replacement string is two &#8208 characters
-    return '<!-- '.str_replace('--', '‐‐', $item).' -->';
+    return '<!-- '.str_replace('--', '- - ', $item).' -->';
 }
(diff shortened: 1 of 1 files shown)


makss:
This patch can affect the output trace log. The trace log encountered the double dash and do not want to spoil the unicode.  ;-)

Please add to your patch a small addition:
The file `/textpattern/lib/txplib_misc.php` line 5702
```
        echo n.comment($msg.n).n;
```

replace to
```
        echo n.'<!-- '.str_replace('-->', '&shy;&shy;>', $msg.n).' -->'.n;
```

vanmelick:
It should affect the output trace log. Having '--' inside a comment is not desired.

makss:
Thank you, I understand.

I've seen double dashes in sql queries, wondering what will happen if we copy a sql query from the trace log and run on mysql server.

vanmelick:
Yes, that won't work.
An alternative could be replacing `--` with `- - ` (dash/space/dash/space). Or even replacing single dashes with dash/space.

vanmelick:
Changed the pull request to replace '--' with '- - ' instead. While this still doesn't enable you to just copy paste an SQL query from the debug info in the HTML comments, it does make it less confusing than replacing '--' with two UTF-8 chars that closely resembled '--'.

Opinions?

philwareham:
@vanmelick @Bloke @rwetzlmayr any feedback on this PR?

rwetzlmayr:
Fine for me. As we've already [lost](https://github.com/textpattern/textpattern/commit/8cd52217349fbb2bab6efb424c3390748d0d203b#diff-4a87f753dc84d7079db6a0dcb04859e7R562) the [Kajagogoo pun](https://www.youtube.com/watch?v=qsjzcT1QKYs&feature=youtu.be&t=1m0s), @vanmelick's markup is just as nice ;)

PR #533, opened 2015-08-22
update tagbuilder after changes in comments_form tags

vanmelick:
A lot of attributes were not available through the tagbuilder, so I only moved the ones that were already there so the tagbuilder at least doesn't create tags with deprecated attributes (due to fixed issue #29).

Is the tagbuilder scheduled for removal or is there still commitment to keep it in TXP?


Bloke:
Thanks Ruud. The tagbuilder is staying for the foreseeable future, but if I have my way it'll be tucked away from the main UI and only take up space if it's required by the administrator.

PR #534, opened 2015-08-25
Add container ability to comment tags that have form attribute

vanmelick:
This affects the following tags: 
* popup_comments
* comments_form
* comments
* comments_preview

If this pull request is accepted, an issue should be opened on the documentation repository to document this change and the fact that we have a popup_comments tag (not documented yet).


philwareham:
Thanks Ruud, this is helpful.

Issue #519, opened 2015-08-15
Use utf8mb4_unicode_ci on MySQL 5.5.3+


vanmelick:
utf8_unicode_ci (and utf8mb4_unicode_ci) is more accurate when sorting compared to utf8_general_ci (and utf8mb4_general_ci), but is around 10% slower.
utf8mb4_unicode_ci allows the full utf8 character set (with a small speed penalty), while utf8_unicode_ci only allows characters that take up to 3 bytes,

* http://stackoverflow.com/questions/766809/whats-the-difference-between-utf8-general-ci-and-utf8-unicode-ci
* https://dev.mysql.com/doc/refman/5.5/en/charset-unicode-utf8mb4.html
* https://dev.mysql.com/doc/refman/5.5/en/charset-unicode-upgrading.html
* https://www.drupal.org/node/2473301
* https://make.wordpress.org/core/2015/04/02/the-utf8mb4-upgrade/

Should we use utf8mb4_unicode_ci collation for *new*  TXP installs on MySQL 5.5.3+... or even require MySQL 5.5.3+ as the minimum version for TXP 4.6?

vanmelick:
I have a patch for this. 
One thing I did encounter is that changing from utf8 to utf8mb4 requires us to reduce the size of some columns. If there's a KEY (whether it be primary, unique or just a regular key) set on a table, the combined size of such a key may not exceed 1000 bytes. This means that if you use a varchar column as a key, that column can be at most 250 characters, because each character can take up to 4 bytes.

I'd recommend reducing varchar(255) columns to varchar(250). And even then there's one case where even that isn't enough... in the txp_prefs table, we have a key that consists of the username, the pref name and the pref id, which are respectively 255 chars, 64 chars and 4 bytes (int) long. Combined that's 1280 bytes. Can be fixed by reducing the pref name to 200 chars, the username to 40 chars.

vanmelick:
Ideally the above only applies to new installs, but because we'd also have to change some of the update scripts, it could also hurt people that are not yet 100% up-to-date, causing a few older upgrade scripts to trigger (scripts that would have to be modified to reduce column sizes).

makss:
If you do add utf8mb4 for a new installation, it is only optional.

Personally, for myself, I do not see any sense to go to this encoding. While I see only the negative side:
* Slowly
* Takes up more space
* Not all versions libmysqlclient support it
[...]

PR #539, opened 2015-08-29
Default to charset utf8mb4

vanmelick:
The MySQL charset utf8 does now allow the full UTF-8 character set. It only allows characters of up to 3 bytes long. utf8mb4 allows the full range of UTF-8 characters (incl 4 byte chars).

This is also a good time to change the collation from generic_ci to unicode_ci. Although slightly slower, it does provide more accurate sorting according to unicode rules.

Wordpress and Drupal have already made these changes.

### DIFF
--- a/textpattern/setup/index.php
+++ b/textpattern/setup/index.php
@@ -382,17 +382,23 @@ function printConfig()
         exit;
     }
 
-    // On 4.1 or greater use UTF-8-tables.
-    $version = mysqli_get_server_info($mylink);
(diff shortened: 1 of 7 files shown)


vanmelick:
This would fix issue #519

philwareham:
@rwetzlmayr and @Bloke - is this PR OK to pull in now?

rwetzlmayr:
I'll have a look soonish.

Issue #547, opened 2015-09-18
Convert trace_add and similar functions to a class


vanmelick:
This would get rid of several globals and probably results in nicer code than what we're currently doing in the trace_log function. Something like this perhaps:

```php
class Trace {
  private static boolean $quiet = false;
  private int $indentLevel = 0;
  private int $timeLast = 0;
  private int $maxMemUsage = 0;
  private int $microstart = getmicrotime();
  private array $pluginCallback = null;
  private array $trace = array();
  public function __construct();
  public static void function setQuiet(boolean $bool);
  public void function start(string $msg); // begin partial measurement
  public void function stop(string $msg = null); // end partial measurement
  public void function log(string $msg); // insert single line in log
  public string function summary();
  public string function result();
}
```
Where we do setQuiet(true) in live mode and only call getTrace() in debug mode.

Comments please. I can implement this if this proposal is accepted.

PR #549, opened 2015-09-21
Convert trace functions to class

vanmelick:
This isn't a pull request yet. It's working though (mostly). Changes are:
* get rid of call back events (don't know how to insert this properly in OO code)
* get rid of displaying plugin callback handlers (don't belong in a trace)
* get rid of a lot of trace related globals, leaving only $trace;
* get rid of the display of memory usage, because it doesn't do it's supposed to do. Instead show only where memory usage peaked.
* different method of showing times in the trace. When nesting, it shows how long each level of nesting took to execute.

Is this is the direction in which we're headed (converting stuff from procedural to OO)?
I'm asking because I planned on doing the same for the parser and for that I needed to do the trace code first (issue #547).

Feedback on code quality is appreciated. This is the first class I've written :)


philwareham:
@vanmelick @rwetzlmayr @Bloke again, any feedback on this PR? Is it OK to merge?

Bloke:
Haven't looked at the code in detail yet (sorry), but it looked good. When I last had a glance through it, the only question I had was that if it's being converted to a class, shouldn't it live in `/vendors` somewhere so it benefits from the class loader?

But I wasn't sure if a trace function is exempt from this scheme since it's low level functionality.

philwareham:
@vanmelick any feedback on what Stef said? Is this ok to merge still?

vanmelick:
It could live in vendors and/or be autoloaded, but then you couldn't trace the autoloading part. I don't know how important it is to trace that part.

Before this is merged, someone should really take a closer look, because it does remove some trace functionality. For example: no more callbacks in the trace functions itself and it doesn't show plugin callbacks.

Bloke:
You're right. It's better outside the vendors folder. I'll apply this locally and give it a test, thanks.

Bloke:
@vanmelick Interesting.

To my untrained eye, it looks as if the values in `Duration` at the same indentation level are the totals for the indented group below it. So, for the particularly slow refresh below, Time (4087.73) + Duration (1758.98) takes you to (roughly) the `[SQL: SELECT name, val FROM txp_prefs WHERE prefs_id = 1 AND user_name = '' ]` row, give or take 10us for processing / display:

```
<!-- Trace log:
  Time(ms) | Duration | Trace
   4087.73 |  1758.98 | [PHP includes]
   4110.44 |          |     [Textpattern autoload dir: 'vendors']
   4110.47 |          |     [Textpattern autoload dir: 'lib']
   4147.19 |   486.84 |     [Load: 'vendors/Txp.php']
   4148.37 |   485.62 |         [Load: 'vendors/Textpattern/Container/FactoryInterface.php']
   4634.02 |          |         [Class loaded: 'Txp']
   4634.10 |  1045.37 |     [Load: 'vendors/Textpattern/Container/Container.php']
   5415.56 |   263.88 |         [Load: 'vendors/Textpattern/Container/ContainerInterface.php']
   5679.47 |          |         [Class loaded: 'Container']
   5679.55 |   123.13 |     [Load: 'vendors/Textpattern/Tag/Registry.php']
   5778.25 |    24.39 |         [Load: 'vendors/Textpattern/Container/ReusableInterface.php']
   5802.67 |          |         [Class loaded: 'Registry']
   5846.81 |   204.96 | [SQL: SELECT name, val FROM txp_prefs WHERE prefs_id = 1 AND user_name = '' ]
   6053.83 |   324.21 | [SQL: SELECT name, data FROM txp_lang WHERE lang = 'en-gb' AND name != '' AND event IN ('public','common') ]
   6378.80 |   175.11 | [Load: 'vendors/Textpattern/Date/Timezone.php']
   6553.90 |          |     [Class loaded: 'Timezone']
   6554.12 |   201.92 | [Loading plugins]
```

At first glance, it's not super obvious that's what's going on, but to retain the group tallies I can see it would be a challenge to represent the values any other way. To some degree, the cumulative times of the previous trace functionality were more intuitive, as you could just quickly scan the column looking for big values to find out what was causing any slowdowns. Now it's a bit more involved to find where bottlenecks lie, but your version gives more detail, which I like. Guess it'll just take some getting used to the layout.

One other thing: does the asterisk represent the single item that contributed to the (peak?) memory usage displayed in the trace summary, or is that the point at which memory usage peaked? In my case the asterisk always seems to be adjacent to the last plugin loaded, so I'm assuming the latter.

Again, the cumulative memory usage from the old trace log allowed us to easily spot memory-hungry plugins, tags or loops, but from your commit log it seems that wasn't reliable so I can see why you changed it. I'd rather have the accuracy than be led to believe that a particular tag was causing memory problems when it wasn't. That said, in my previous experience with the trace log, I used it to quickly diagnose that an old version of smd_bio was being terribly inefficient (hogging memory and taking ages to run its queries) which led to a series of cacheing improvements that sped up page render times by over 85%!

vanmelick:
The asterisks indicates where peak memory usage occurs. Sometimes it shows two asterisks, meaning it peaked somewhere in between.

The old tracelog showed memory usage just before (and after) parsing a tag. The increases you saw then were mainly due to the string output of tags as they are parsed and increased use of globals. What it didn't show was how memory usage behaved while parsing the tag. You could have a plugin that allocated and released a GB of RAM without seeing it in the tag trace.

Perhaps we should also ask forum crowd for input.

Bloke:
I've merged it for now, but we should probably reinstate any callbacks at some point. For reference, you register callbacks inside a class using array syntax:

    register_callback(array(__CLASS__, 'localFunction'), 'hook');

Raising callback_events are probably very similar, though I've never done it.

Bloke:
Hmmm, some plugins use `include_once publish.php` in order to gain access to the article parsing functions, etc. Upon doing so:

    Notice "Undefined variable: trace" in textpattern/publish.php at line 51.

Any way round this? Adding `global $trace` to the file maybe?

vanmelick:
You'd also have to add a $trace->stop() to /index.php and a $trace->start() to textpattern/publish.php, because the stop() method should always be preceded by a start() method.

Bloke:
Cool. Is that something you fancy adding as part of this patch?

jools-r:
I immediately got some warning messages from some plugins that use `trace_add`. Can you provide some infos (e.g. on the forum) on the correct way to replace `trace_add` functions in existing plugins.

I tried adding `global $trace;` to the plugin function and then replaced

`trace_add(…)` with `$trace->log(…)`. Correct?

vanmelick:
@jools-r: http://forum.textpattern.com/viewtopic.php?pid=298008

jools-r:
Perfect. Thanks!

Issue #557, opened 2015-10-16
Remove broken Import functionality


philwareham:
The import function doesn't actually work any more for WordPress, and the other CMSes we support importing from are effectively dead.

As discussed at our recent meetup, we should ~~deprecate~~ delete this in v4.6 and hide it from admin panels. Another thing that we don't need to support in future (not that is has been supported from some time).

~~Potentially remove all import code in a future release? Not sure how we handle deprecated code in this project - only at major release numbers (i.e. 5.0 - bearing in mind e've never had a major release number since initial release) or can it be done at point releases (i.e. 4.7).~~

If anyone feels an import/export feature would be beneficial in future then that can be written from scratch.

vanmelick:
If it doesn't work, removing it completely won't affect anyone. And those that need it to convert from an ancient obscure CMS can download an older TXP version to do the import and then upgrade.

I'd skip the deprecating phase. This is something you'd only use directly after installing.

+1 on removing it completely in 4.6.

Bloke:
I'd also be tempted to just bin it. It doesn't work and most of the supported systems don't even exist any more.

philwareham:
Great! The less code the better in my opinion. Let's get rid. @vanmelick, would you be willing to do that for us? (I've already assigned plenty of tasks to Stef!)

vanmelick:
Sure. I'll prepare a patch.

philwareham:
Fone

PR #559, opened 2015-10-16
Remove broken import functionality

vanmelick:
Fix issue #557


philwareham:
The language strings will needs to keep the import strings for now, so please revert that part of the change.

FYI the language text files we are planning to bundle into v4.7 release, so we will redux them at that point.

vanmelick:
Do you mean en-gb.php AND en-gb.txt or just one of these?

philwareham:
Both of those need to left untouched for now. Thanks!

vanmelick:
Lang file changes are reverted.

philwareham:
Thanks Ruud!

Issue #556, opened 2015-10-16
Use native PHP namespaces


rwetzlmayr:
Textpattern currently employs pseudo-namespacing with underscores, e.g. ``Textpattern_Textfilter_Textile``, mainly serving as a workaround for pre-PHP-5.3 compatibility which is not needed any more. 

I propose to switch to PHP's native namespaces before the stable release.

Bloke:
Sounds like a plan.

rwetzlmayr:
TIL that parts of the framework @gocom started to build is apparently dead code not used anywhere in the project.

1. \Textpattern\Filter
1. \Textpattern\Request
1. TBC

rwetzlmayr:
Needs testing.

philwareham:
I'm getting this error when I try to view the preferences panel, is it related to this patch?:

    Fatal error: Class 'Textpattern\Date\DateTimeZone' not found in /Users/pwareham/Sites/textpattern/textpattern/vendors/Textpattern/Date/Timezone.php on line 118

Bloke:
@philwareham  Fixed in https://github.com/textpattern/textpattern/commit/dab4561b6a518492192445484e4e1d2dc2f69db2

vanmelick:
This bit of code in Textpattern/Loader.php confuses me:
``` php
if ($this->namespace !== null && strpos($class, $this->namespace.$this->separator) !== 0 ||
    !preg_match('/^[\\a-zA-Z_\x7f-\xff][a-zA-Z0-9_\\\x7f-\xff]*$/', $class)
```
That regex doesn't just match the class, it matches namespace + class. Shouldn't the regex be like this:

``` php
^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*(\\[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*)*$
```

rwetzlmayr:
Considered `Textpattern\BuilderTags` as a refactoring candidate. But as this class is such an outlier with no consumer code outside its own little space I left it as it is.

rwetzlmayr:
> https://bugs.php.net/bug.php?id=50731

@vanmelick: The current regex matches names with a leading backslash or leading alphas AFAICS. I suspect that this is related to https://bugs.php.net/bug.php?id=50731. 

[...]

PR #562, opened 2015-10-21
Remove support for underscore pseudo-namespacing

vanmelick:
Completes work done in #556 by removing the ability to use underscores for namespacing.


rwetzlmayr:
Thanks.  I wonder if we could replace our homebrewn loader with the one from Composer. One less wheel to  re-invent...

Has anyone investigated?

Issue #567, opened 2015-10-23
SQL coding style and updates


vanmelick:
Proposals:

1. Write all reserved SQL keywords as uppercase (consistency).
2. Remove backticks around table/column names where not needed (consistency)
3. When inserting forms/pages/articles/css (in txpsql.php), use heredocs instead of unreadable long lines.
4. On new installs, don't trigger the various update scripts, but let txpsql.php handle it completely. Why?:
 * If the goal was to let txpsql.php be the same as in the most ancient TXP version, then that isn't true anyway at the moment. Some newer stuff is in there as well and has been for quite some time.
 * If you currently want to know what the up-to-date DB structure is, you either have to put together all the pieces from txpsql.php and the various update scripts or you have to query the database itself (and assume you didn't change anything yourself). Having the end result in txpsql.php would be self-documenting; it's immediately clear what the table structure should be.
 * No more creating tables that are removed in later update scripts or changes that were reverted in later updates.

(and yes, I'm volunteering to do this).

Bloke:
Yes please!

The update scripts are a royal pain anyway and could do with some try... catch loveliness at some point so they can abort if something doesn't go to plan. That way, upgrades can continue where they left of after the sticking point has been corrected without the database or file system  being in an unknown state.

At the moment, the update scripts blunder through to the end. Unless something truly catastrophic happens, the version number is set to the latest version at the end. Once the version number is set, upgrades cannot be re-run without manually altering the version string in the database, touching a file and logging back in.

For reference: [Jukka's legendary post on the subject](http://forum.textpattern.com/viewtopic.php?pid=272537#p272537).

vanmelick:
5. replace the use of PFX with safe_pfx() calls
6. replace ```key=value, key2=value2``` with ```key = value, key2 = value2``` (consistency)

Is proposal 6 okay as well... or should we remove those spaces everywhere?

Bloke:
I don't mind either way, but personally I find spaces more readable. By all means go for it.

vanmelick:
Okay, first batch of changes sent in pull request.

Next on my TODO list is:
* update the SQL queries in other parts of TXP, which should be just cosmetic changes (uppercase + spacing). A lot of work so that depends on whether the style people like the style in this request.
* removing update files that don't contain content.
[...]

PR #568, opened 2015-10-27
SQL: installation / update scripts

vanmelick:
This is a big one. Changes:
* Consistent layout of SQL queries, with upper case keywords, no backticks unless needed, etc. It should be a lot easier to read the setup / update files now, especially txpsql.php
* setup/txpsql.php can now handle the entire installation part of the database, without having to run update scripts after the first login.
* All the big chunks of text like style sheets, forms and the initial article are now stored in separate external files, which are much easier to edit (and greatly reduces the size of txpsql.php)
* Moved SQL queries from setup/index.php to setup/txpsql.php (which is a more appropriate place)
* Use of built in functions where possible to reduce the amount of code needed during updates.
* If SQL errors happen during installation, show usable debug info which includes the actual queries that caused the errors (in a nice style now as well).
* Remove several things during installation that weren't needed, like display widths on table columns (confusing, really), auto_increment values (those are set automatically), and pack_keys (which MySQL knows how to set better than we do).
* Found a few inconsistencies due to changes in txpsql that were never part of update scripts, causing differences between TXP versions, even if updated to the same version. Fixed.
* Found and fixed a few small bugs in the update scripts.
* The _to_4.5.7 file wasn't called during updates.
* Avoid the missing table error (due to the removal of the ipban table) when using SVN after login.

What I've tested:
* update from 4.0.3, 4.2.0, 4.3.0. Table structures match those of 4.6.0.
* compared the "old" 4.6 branch with this new one. Table structures match.
* compared prefs table from 4.0.3 (updated), 4.6 (old) and 4.6 (this one). They match (minor differences, but those are intentional)

This partially fixes issue #567 (see that issue for more details).


Bloke:
Holy awesome! This is fantastic work, thank you. It's all soooooo much easier to read. We should aim to get this merged into both the `master` and `admin-layout-update` branches quickly before too many other things change around the periphery and make merging trickier.

I won't commit it right now because https://github.com/vanmelick/textpattern/commit/6196a87649a3bc3afe370a6146abd1bec9da5af8 will impact Phil's scripts that build `txpsql.php` so I'll let him review what needs to change there first and test it more fully than I can at this time of night. But I expect it will make the scripts simpler and much easier to manage going forward.

I haven't tried an upgrade from <4.2.0 yet. Does the synchronisation of all the SQL with today's table structures also mean that the `Unknown column 'user_name'` and other related 'known errors' documented in the [README](https://github.com/textpattern/textpattern/blob/master/README.txt) will be a thing of the past and those parts of the README can be edited out?

Thank you again. This is above and beyond the call of duty.

rwetzlmayr:
NB: I think @philwareham uses https://github.com/rwetzlmayr/wet_sql2php to generate the relevant parts of txpsql.php whenever he updates the public theme. SQL formatting is part of it.

philwareham:
Great work Ruud!

That's correct, I use wetsql2php. I will try out the patch and see how that affects me later today.

Whilst on the subject of setup, when I met with Stef the other day we touched on the subject of the default 'what to do next' article that is installed at setup. This is currently only in English which doesn't sit well with our multinational support.

As part of v4.7 we want to decouple the language textpacks from RPC and instead bundle them into the package (this also has the benefit of allowing textpacks to be versioned to their target Textpattern release and for all the crufty old unused strings to be removed), I guess we could also find a method of having the welcome article as a flat text file that is imported to database at install - allowing for multi language versions of it. As I say, it's 4.7 work but worth thinking about now.

vanmelick:
This is how it affects you:

Before:
* 1 file: txpsql.php
* use wetsql2php
* copy/paste contents in txpsql.php (there's no 'save' button?)

After
* txpsql.php and lots of small text files
* copy paste contents of individual changed forms/pages/css into separate text files.

I didn't know wetsql2php existed and assumed it was a lot of work to update txpsql.php manually. If the wetsql2php route is preferred, I can revert this part of the pull request.

vanmelick:
It doesn't fix the pre 4.2.0 errors, but the readme does need an update, because the errors are not as documented (some the same, some different). Perhaps some of the errors can be fixed. I'd have to look into that.

vanmelick:
@philwareham, how do you update txpsql.php when you change the article body/excerpt?

philwareham:
@vanmelick hey, this is great stuff!

In the older system, if I wanted to update the public theme, I had to paste those forms/pages/styles into a working text pattern install, then run wetsql2php, then copy/paste the results into the txpsql.php file. A right slog. This will be much better.

Also, to amend article body/excerpt I would do a similar thing, either run wetsql2php or hand edit the sql.

One thing I would say though, is can the structure within 'setup' directory follow what we are proposing for themes in 4.7 onwards - that way it'll be relatively simply to move this over when the time comes.

@Bloke can probably give more info, but in simple terms the forms, pages and styles are within subdirectories and have a certain naming convention.

Bloke:
I really like the separate files as it's easier to read, and as Phil says leaves the door open for separating out other parts of the installation routines. I'm not at the sharp end of keeping it up-to-date, but I'm sure a suitable workflow can be figured out.

Having the default content separately editable will be a boon for modders too, as they could more easily alter the default content (Forms, for example) prior to a fresh client installation without hacking the previously impenetrable `txpsql.php` file. Granted, I can see people trying to remove files for those Forms they don't use, but that will (rightly) break the installation routine under our 'convention over configuration' model.

Bloke:
Regarding what Phil says about directory structure, public Themes are likely going to be in a format [akin to this](http://forum.textpattern.com/viewtopic.php?pid=291622#p291622). In short:

* Pages, Forms and Styles separated into subdirs.
* Files ending in `.txp`.
* Form files of the format `type.formName.txp`

philwareham:
I have already moved the textpattern-default-theme repo to this directory/naming model in readiness, by the way.

vanmelick:
Questions:
* In which directory does the default theme go?
* Should txpsql automatically detect how many forms/pages/css files there are or can this be hardcoded (assuming the amount of files and names don't change much)?
* Which format to use for the default article body / excerpt? (article.body.txp / article.excerpt.txp?)
* Where do we put the default article body / excerpt (would that be part of the theme as well?)?

TODO: textile article body/excerpt instead of storing them as flat files as well.

Bloke:
`dirname(txpath).DS.'themes'.DS.'default'` I would expect.

rwetzlmayr:
> `dirname(txpath).DS.'themes'.DS.'default'` I would expect.

This would leave us with a ``theme`` (admin themes) and a ``themes`` (public themes) folder in txpath

philwareham:
The article body/excerpt is up for discussion really. I personally don't think articles should be part of a theme, since they are content (although I can see some benefits of themes having their own ready-made articles). I think it should be Textile, the same as the pophelp files already are but others might have reasons why not (maybe to do with Markdown/Textile, don't know).

For now (v4.6) the 'setup' directory is probably the safest place to put the theme. From v4.7 the themes are proposed to have their own 'themes' directory in the site root but if we do that now it might confuse people into thinking theme are available at v4.6?

vanmelick:
I've made a few changes. The /setup dir now contains these files:

./en-gb.php
./css
./css/ie8.css
./css/default.css
./txpsql.php
./setup-langs.php
./pages
./pages/error_default.txp
./pages/archive.txp
./pages/default.txp
./prefs.php
./article.excerpt.txp
./article.body.txp
./forms
./forms/article.default.txp
./forms/misc.images.txp
./forms/comment.comments_display.txp
./forms/misc.search_input.txp
./forms/file.files.txp
./forms/link.plainlinks.txp
./forms/comment.comments.txp
./forms/article.article_listing.txp
./forms/comment.comment_form.txp
./forms/comment.popup_comments.txp
./forms/article.search_results.txp
./index.php

If you add more css/pages/forms files, no change is required in txpsql.php (it reads the directories for files that look like css/pages/forms).

vanmelick:
About 'essential' forms. What surprises me is that the ones you can't delete are those you CAN override when using them, while 'popup_comments' is NOT marked as essential and that's the only one that you cannot change without editing a PHP file.

In the proposed pull request, if a page/css/form file doesn't exist, it isn't inserted in the database.

philwareham:
Looks good to me.

I assume there is no increased security risk using this method than the previous setup files - nobody can inject stuff into your system this way over and above what they could do before (I know people should remove setup folder after install which negates this, but probably some don't)?

vanmelick:
Yes. If you can add extra files to /setup, you can also replace the existing setup and other files. No more risk than before, I think.

vanmelick:
Those were my final changes to the update/install scripts.
Once those are applied, I'll send a pull request for changes to the files outside /setup and /update to get some more consistency in SQL queries there as well. I have those ready to go as well in the same branch, just not committed yet.

philwareham:
Do you want me to merge the pull request now or wait for further commits?

vanmelick:
I'd merge this part first, then I'll issue a new pull request for the rest. 
Otherwise it's a really big chunk all at once.

philwareham:
Thanks for all this work Ruud, will get some users in the forum to test this over the next couple of days.

vanmelick:
You may want to wait. There's more coming :)

PR #570, opened 2015-10-29
SQL coding style fixes

vanmelick:
Similar to the changes made to the update files (#568), but now for the rest of the code base.

* SQL keywords in upper case
* Consisted use of spacing (spaces around = and after ,)
* Consistent quoting style. SQL strings are now all delimited by double quotes, table names in single quotes (improves readability)


philwareham:
This is going to be a tricky merge into the admin-layout-update branch, but I'll see what I can do. A of of code has changed/moved about in that branch (mainly the HTML rendering code).

philwareham:
Tested and all seemed OK in limited test. Thanks Ruud!!

philwareham:
Just noticed that this patch breaks the styling of list column sorters - because the `asc` and `desc` classnames have been changed to `ASC` and `DESC`.

For example:

    class="DESC txp-list-col-title"

Needs fixing because the CSS expects lowercase classnames.

philwareham:
Issue #567 notes this bug.

PR #572, opened 2015-10-30
Fix regression caused by #570 (SQL coding style fixes)

vanmelick:
I could fix it either by lowercasing $dir or by making an exception to the rule of having uppercase SQL keywords. The latter was preferable here because the change in case also affected the contents of a pref, causing a potential reversal of stored sort direction on upgrades.


philwareham:
Thanks a lot Ruud

PR #577, opened 2015-10-31
Parser optimization and removal of ancient code

vanmelick:
* Let $registry->processTags() return false if the tag is not registered or not callable (and a string otherwise). This makes it possible to avoid checking twice if a tag is registered, which makes parsing slightly faster.
* Clarify why doing is_callable during tag registering isn't enough and why we only do a syntax check there. This had me puzzled for a while.
* During processing, if we encounter an unregistered tag and it does appear to be an actual tag, register that tag and then process it again. It's enough to get one "unregistered tag" warning for each tag (if a tag occurs more than once).
* Remove some ancient code that has long been deprecated. No point in deprecating if the code never gets removed.

### DIFF
--- a/textpattern/lib/txplib_publish.php
+++ b/textpattern/lib/txplib_publish.php
@@ -458,23 +458,16 @@ function processTags($tag, $atts, $thing = null)
         $registry = Txp::get('\Textpattern\Tag\Registry');
     }
 
-    if ($registry->isRegistered($tag)) {
-        $out = $registry->process($tag, splat($atts), $thing);
-    }
+    $out = $registry->process($tag, splat($atts), $thing);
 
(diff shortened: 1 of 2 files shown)


vanmelick:
Btw, if the goal of registering tags is to make Textpattern more secure, should we even provide a fallback for non-registered tags?

rwetzlmayr:
> should we even provide a fallback for non-registered tags? 

As 4.6.0 will be the first version which requires tag registering I would consider it a courtesy to be a bit liberal about what we accept as a tag. It will give plugin authors more time to get their custom tags in shape and avoid to break sites upon updates. We can think of dropping the fallback at later versions.

philwareham:
I've been using this patch for a while without any errors, so I will merge.

PR #594, opened 2015-12-03
Admin layout update

philwareham:
Merge that new admin layout into the master, finally. Any other admin layout amends can be done as part of the normal master dev work.


petecooper:
Fantastic work, fellas. Well done!

rwetzlmayr:
Marvellous work! Out of curiosity: Why is "Create new link" styled as a button while "Create new form" is a link w/ an icon?

screenshot - 07 12 2015 15_55_47

__vs.__

screenshot - 07 12 2015 15_55_40

philwareham:
@rwetzlmayr Thanks!

The 'create new form' follows the pattern used on the article write page. I guess it could be a button instead - works either way really.

PR #605, opened 2015-12-23
Make columns default to CURRENT_TIMESTAMP or CURDATE() if mysql >= 5 (fixes #591)

### LINKED ISSUE #591: MySQL 5.7: default SQL mode conflicts with having 0000-00-00 00:00:00 as default datetime
See:
* http://forum.textpattern.com/viewtopic.php?id=46577
* http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sql-mode-setting
* http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_no_zero_date

If the MySQL configuration setting ```sql-mode``` contains ```STRICT_ALL_TABLES``` and/or ```NO_ZERO_DATE``` (possible also if it contains ```STRICT_TRANS_TABLES```), creating tables that have a ```datetime``` column with default value set to ```0000-00-00 00:00:00``` fails.

In MySQL 5.7, ```sql-mode``` contains  ```STRICT_TRANS_TABLES``` and ```NO_ZERO_DATE``` by default, causing errors while setting up the tables:
1. Invalid default value for ‘Posted’
3. Invalid default value for ‘posted’
5. Invalid default value for ‘date_banned’
6. Invalid default value for ‘issue_time’
7. Invalid default value for ‘date’
8. Invalid default value for ‘date’
15. Invalid default value for ‘time’
16. Invalid default value for ‘last_access’

IIRC, the value "0000-00-00 00:00:00" has special meaning for the expired column, so we'd have to find an alternative for that as well as setting a different default value or not setting a default at all for other affected datetime columns.

Another solution might be to override the SQL mode in TXP before issuing any table related queries:
```SET SESSION sql_mode = '';```

walker:
I think that this should fix #591.

I set it to use triggers for mySQL versions 5 or greater, as that's when they were introduced.

### DIFF
--- a/textpattern/setup/txpsql.php
+++ b/textpattern/setup/txpsql.php
@@ -106,12 +106,16 @@ $themedir = txpath.DS.'setup';
 
 $create_sql = array();
 
+$mysql_version = (float) mysqli_get_server_info($link);
+# If we can't add the trigger (MySQL is < 5), then let's set that DEFAULT as previously specified
+$feed_time_default = ($mysql_version>=5) ? '' : ' DEFAULT "0000-00-00"';
+
 $create_sql[] = "CREATE TABLE `".PFX."textpattern` (
     ID              INT          NOT NULL AUTO_INCREMENT,
-    Posted          DATETIME     NOT NULL DEFAULT '0000-00-00 00:00:00',
-    Expires         DATETIME     NOT NULL DEFAULT '0000-00-00 00:00:00',
+    Posted          DATETIME     NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    Expires         DATETIME     NOT NULL DEFAULT CURRENT_TIMESTAMP,
     AuthorID        VARCHAR(64)  NOT NULL DEFAULT '',
-    LastMod         DATETIME     NOT NULL DEFAULT '0000-00-00 00:00:00',
+    LastMod         DATETIME     NOT NULL DEFAULT CURRENT_TIMESTAMP,
     LastModID       VARCHAR(64)  NOT NULL DEFAULT '',
     Title           VARCHAR(255) NOT NULL DEFAULT '',
     Title_html      VARCHAR(255) NOT NULL DEFAULT '',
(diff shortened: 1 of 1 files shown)


philwareham:
Thank you for your email.

I will be out of the office until 4th January 2016 and may not be able to respond to emails. I will reply to your message as soon as possible on my return.

vanmelick:
A few comments:
* TXP requires MySQL versions above 5, so there's no need to check for that.
* Expires should not default to CURRENT_TIMESTAMP (not a sensible default) and there should be a way to indicate that an article does not have an expire date set, so that column shouldn't have NOT NULL.
* last_access should not default to CURRENT_TIMESTAMP. Currently the zero time is used to indicate the user has never logged in. This can be achieved by removing the NOT NULL limitation.
* @$lastmod = date('Y-m-d H:i:s');@ for the language strings may have undesired side effects. It might prevent language updates from showing up, until a new update becomes available after TXP is installed. I'm not familiar enough with the language code to be 100% sure.
* The patch doesn't prevent TXP from inserting zero dates (and TXP does do that).
* The patch doesn't convert tables in existing installs.
* I'm tempted to say that in all the other cases the use of CURRENT_TIMESTAMP is the right thing to do. What makes me hesitate is the 4 open timezone issues, which might result in moving the setting of timestamps entirely to TXP instead of relying on MySQL to do this. Setting a default in the table would then give the wrong signal. Since TXP already explicitly sets the timestamp in all (?) cases, I'd opt to remove the default entirely. For the same reason, the trigger isn't needed (feedtime is set explicitly during insert)

walker:
Augh. You have, of course, exposed the underlying complexities that I either ignored or was unaware of.

Perhaps I could better focus time on the open timezone issues?

vanmelick:
I'll prepare a new patch this weekend.

vanmelick:
While looking into this a bit further, I found that using CURRENT_TIMESTAMP as default value for DATETIME columns wasn't supported until MySQL 5.6.5. Surprising. Anyway, new patch sent via pull request #612

philwareham:
Closing as this has been fixed in another pull request. Thanks guys, appreciated.

PR #579, opened 2015-11-01
Update scripts: less chance of errors / abort on error

vanmelick:
See commit messages for details. Goal is to fix issue #573

PR #612, opened 2016-01-09
No zero dates in SQL. Category name 64 chars. Re-create txp_token

vanmelick:
MySQL 5.7 by default no longer allows 0000-00-00 dates, which were used as the default on several columns in TXP tables.
This patch removes those defaults for NOT NULL columns and sets the default to NULL for NULL columns (textpattern:Expires and txp_users:last_access)
Zero dates in existing rows are converted to sensible values that are allowed.
This fixes issue #591 

Categories, when created or edited can be 64 chars long. A few other tables that used these categories allowed for longer names. That doesn't make sense, because they can't be longer than 64 chars, so I've changed those colums to varchar(64).

txp_token was added to the update scripts, but without NOT NULL and it was also missing from txpsql.php. Added it there and recreated the table to be more in line with the other tables.

### DIFF
--- a/textpattern/lib/constants.php
+++ b/textpattern/lib/constants.php
@@ -183,7 +183,7 @@ if (!defined('REGEXP_UTF8')) {
  * @package DB
  */
 
-define('NULLDATETIME', '\'0000-00-00 00:00:00\'');
+define('NULLDATETIME', 'NULL');
 
 /**
  * Permlink URL mode.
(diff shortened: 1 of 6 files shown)


vanmelick:
PS. If this is merged, it'll probably conflict with pull request #579. If so, let me know and I'll fix the merge conflicts.

philwareham:
Thanks Ruud!

philwareham:
@vanmelick Seems there is an issue with this pull request, see here for more info: http://forum.textpattern.com/viewtopic.php?id=46706

I can either leave in place if a quick fix is possible, or revert. Let me know. Cheers!

PR #613, opened 2016-01-11
Fix NULL regressions

vanmelick:
Fix a few regressions caused by my earlier patch which removed zero dates.
This now does affect plugins that use the Expired column in the textpattern table. I removed the NULLDATETIME constant, because if you change it from NULL to something else, you also have to modify a lot of queries. No point in having it as a constant anymore.

PS. Please add "*~" to .gitignore (to exclude backup files created by unix text editors.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ ehthumbs.db
 Thumbs.db
 /vendor
 /composer.lock
+*~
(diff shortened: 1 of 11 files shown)


philwareham:
Thanks, I'll let people know in the forum.

PR #615, opened 2016-01-19
Update checksums and history

vanmelick:


philwareham:
Thanks Ruud, also for the history update.

PR #621, opened 2016-02-08
Set theme to Hive when upgrading (fix #607)

### LINKED ISSUE #607: Active admin theme needs to be reset to 'Hive' for 4.6
I'm not personally going to be updating Classic or Remora admin themes for Textpattern 4.6 (at least not in the short-term - but probably never). I don't have the time or desire to support multiple admin themes as part of the official distribution.

Therefore, on upgrade, the system should reset the active admin theme to 'Hive' regardless of what admin themes are installed previously. If a user then wants to switch to another theme they have, then of course they can, after upgrade.

I will offset the pain of this by also supplying a neutral colour scheme version of Hive for those who don't like the yellow look.

So, it'll be:

* Hive (Classic Yellow) - which would be the default choice on install, and the one selected after a 4.6 upgrade.
* Hive (Neutral Grey)

Of course, if someone else wants to take ownership of Classic and Remora themes before 4.6 release, they are welcome to. I've not moved them to Sass or updated them for the new admin structure past what was done at v4.5.

vanmelick:
fix #607


philwareham:
Great! Thanks Ruud, one less issue holding up 4.6 release.

Issue #618, opened 2016-01-31
User_Error "Field 'LastMod' doesn't have a default value" when duplicating an article


phiw13:
Articles panel: select an article (checkbox), using the widget at the bottom of the table, choose duplicate.

```
User_Error "Field 'LastMod' doesn't have a default value"
in /Users/<username>/Sites/site/textpattern/lib/txplib_db.php at line 401.
adminErrorHandler()
textpattern/lib/txplib_db.php:401 trigger_error()
textpattern/lib/txplib_db.php:477 safe_query()
textpattern/include/txp_list.php:710 safe_insert()
textpattern/include/txp_list.php:58 list_multi_edit()
textpattern/index.php:210 include()
```

And afaict, the article is not duplicated.
Tested with an old site( upgrade to 4.6dev latest), a new site (created with 4.6dev), with old and newly created articles.

MySQL: 5.6.21, PHP: 5.5.30

vanmelick:
I cannot reproduce this, but I'm on MySQL 5.5 and this may be something that happens only in 5.6+.
Probably it's because of line 702 in txp_list.php, where LastMod is unset in the array that is used to insert data for the duplicated article into the database.

vanmelick:
The easiest fix is to not unset LastMod. The effect would be that the copied article has the same last modified timestamp as the original, which would match the behaviour of duplicating the article from the write tab. This is related to issue #614, so once a decision has been reached there on how to solve it, I'll tackle both issues with a single patch. I'd pick the "write tab" duplicating behaviour over the "article list tab" approach.

phiw13:
> I'd pick the "write tab" duplicating behaviour over the "article list tab" approach.

Yeah, I would expect both “duplicate” functions to return the same results.

philwareham:
Fix now, thanks Ruud.

phiw13:
Hmm, I still get an error when duplicating an article from the Articles panel (and no article is duplicated, AFAICS):

```
User_Error "Incorrect datetime value: '' for column 'Expires' at row 1"
in /Users/username/Sites/_txp/textpattern/lib/txplib_db.php at line 401.
adminErrorHandler()
textpattern/lib/txplib_db.php:401 trigger_error()
textpattern/lib/txplib_db.php:477 safe_query()
textpattern/include/txp_list.php:712 safe_insert()
textpattern/include/txp_list.php:58 list_multi_edit()
textpattern/index.php:210 include()
```

[...]

Issue #614, opened 2016-01-16
Mismatch with published date for copied article


cara-tm:
Latest 4.6 dev branch.
Copy an article; see the publish date: in article list context, this is the actual publication date which is displayed but in individual article context this is the parent article's date which is displayed.

vanmelick:
There are more differences:
Article list => select article => duplicate: 
* posted set to current time
* modified set to current time
* status same as the original article.

Article edit window => duplicate:
* posted set identical to original article posted time
* modified set identical to original article modified time
* status set to draft.

Are these differences intentional? If not, which of these is the correct behaviour?

Bloke:
@cara-tm Sorry, not sure I follow. If I duplicate an article from the Write panel, the dates stay the same as the original. If I then go and view the article on the front side (in either List or Individual Article context), the date is correct. If I edit either article's Posted date and view it in either context, the date is as I set it. What am I missing?

@vanmelick Yes, intentional(ish), but could probably do with some tweaking. When duplicating an individual article from the Write panel, the Status should definitely become Draft, as the workflow is "I'm basing a new article on this old one" and you won't want it to appear immediately on the site if it's an exact clone. The dates, well, that's open to interpretation. One could argue that the dates should be changed to 'now'. For a blog-style workflow, that may be desirable so your new content (when it's made Live) appears as most recent. But for a corporate-style site where you've manually set the dates so your articles appear in a particular order, you might want the dates to remain the same. Although you can change to now with a single click on the _reset time to now_ checkbox.

From that perspective, the fact that you can choose how the dates are altered:

* one by one: clone from the Write panel, get an exact clone.
* en-masse from the Articles panel: all selected articles get 'now' date) gives flexibility.

But it could be confusing.

In either case, the fact the status flags stay the same when cloning from the Articles panel might be construed as undesirable.

But it goes further than that: does the change of status only affect Live articles? i.e. if Live, drop to Draft in all cases, but if Pending/Hidden, leave alone? And what about Sticky articles? Leave those too? Make them Draft?

I'm not sure of the exact answers. But you're right that we should either standardise both panels, or make them distinct in some way while making it clear what the outcome will be.

Open to discussion and patches on this.

vanmelick:
I can't think of a situation where you'd want to copy a live article and have it go live without changing it, let alone multiple articles all at once. The "duplicate" function from the article edit window makes more sense to me in how it functions. A simple patch could be to just remove it as a multi-edit option.

[...]

PR #624, opened 2016-02-13
More consistent duplicate behaviour

vanmelick:
Fixes issue #614. Changes are:

Duplicate from article write tab:
* reset posted time if that option is checked.
* always set last modified time to NOW.

Duplicate from articles list tab:
* don't change Posted and Expires dates.
* reset comments_count to zero.
* set last modified time to NOW (fixes issue #618).
* set status to draft if the old status was live/sticky

One difference remains. On the write tab, the url_title is regenerated from the title when duplicating. On the articles list tab, this doesn't happen.

Side effect: for new articles, the last modified time is now always set to NOW(). Previously, for articles set in the future it was set to the posted date. However when editing such an article, it would be reset to NOW(), so that didn't make much sense.

### DIFF
--- a/textpattern/include/txp_article.php
+++ b/textpattern/include/txp_article.php
@@ -163,7 +163,7 @@ function article_post()
         $Annotate = (int) $Annotate;
 
         // Set and validate article timestamp.
-        if ($publish_now == 1) {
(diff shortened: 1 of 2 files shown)


philwareham:
Tested and seems good. Merged, thanks.

Issue #603, opened 2015-12-20
Fatal error: Uncaught Exception: invalid_argument: locale


vanmelick:
I got this error after I installed the nl-nl language and then selected that language at the top in the languages tab.
```
Fatal error: Uncaught Exception: invalid_argument: locale
in /var/www/html/textpattern/textpattern/vendors/Textpattern/L10n/Locale.php:141
Stack trace: 
#0 /var/www/html/textpattern/textpattern/include/txp_lang.php(310): Textpattern\L10n\Locale->setLocale(6, 'nl-nl')
#1 /var/www/html/textpattern/textpattern/include/txp_lang.php(49): save_language()
#2 /var/www/html/textpattern/textpattern/index.php(210): include('/var/www/html/t...')
#3 {main} thrown in /var/www/html/textpattern/textpattern/vendors/Textpattern/L10n/Locale.php on line 141
```
This occurs on a server with the following locales:
```
# locale -a
C
C.UTF-8
POSIX
en_US.utf8
```
Changing the language in TXP to en-gb or en-us did not cause any errors. 

There's two issues here, I think:
1. The exception isn't caught.
2. Why was the exception thrown? Changing the language to one I just installed seems like a valid thing to do.

phiw13:
See also this forum thread: http://forum.textpattern.com/viewtopic.php?id=40685

philwareham:
Fixed. Closing.

brachycera:
Heads up for anyone else who has Problems updating from TXP 4.5.7 to TXP 4.6.0. I had the following Error:

`Fatal error: Uncaught exception 'Exception' with message 'invalid_argument: locale' in PATH_TO/textpattern/vendors/Textpattern/L10n/Locale.php on line 141`

I had two `<txp:posted>` Tags with the Attribute value for `lang="de"`. After changing the value to `lang="de_DE"` the Error went away. This Error also occurred  with other languages.

Bloke:
I'm looking into this, thanks. I think all we need to do is add judicial use of `try... catch` blocks around any `setLocale()` calls that might try to set the locale to anything other than your default. I just need to set up a test environment to simulate it and I'll see if my patch works.

For your specific issue (inside the `safe_strftime()` function) I've done this as an as-yet-untested kludge, replacing the block from line 3111 with this:

```
    if ($override_locale) {
        $oldLocale = Txp::get('\Textpattern\L10n\Locale')->getLocale(LC_TIME);

        try {
[...]

PR #625, opened 2016-02-14
Fallback to C locale if desired locale doesn't exist. Fixes issue #603

vanmelick:
C locale always exists, so this avoids an exception getting thrown. Not falling back to C locale would mean the previous locale would remain set, which is undesirable, because then there would be a weird mismatch between locale and language.

### DIFF
--- a/textpattern/include/txp_lang.php
+++ b/textpattern/include/txp_lang.php
@@ -306,12 +306,17 @@ function save_language()
     )));
 
     if (safe_field("lang", 'txp_lang', "lang = '".doSlash($language)."' LIMIT 1")) {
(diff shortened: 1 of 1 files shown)


vanmelick:
Please note that this pull request adds a language string, which is shown when you choose a language for which no matching locale is installed, so you don't get proper localisation, defaulting to locale C:
gTxt('locale_not_available_for_language', array('{name}' => $language))

philwareham:
Can you please provide the full English example of what that language string should say? Thanks.

vanmelick:
"No locale available for language {name}. Using default locale instead."

Something like that. Would you like me to change {name} to {language} or {lang} or is {name} good enough?

philwareham:
I think `{name}` is fine in this instance.

Issue #623, opened 2016-02-12
Comment woes: <txp:comments_error /> returns nothing


phiw13:
STR:
* start filling a comment form, but leave (at least) one required field blank.
* press ‘preview’

Actual: the preview page appears with a preview of the non-blank fields, the blank required field highlighted. There is however **no** message telling the user the type of error.

expected: a message should appear explain the user error (see the behaviour with TXP 4.5.x.

(tested with the default form template shipped by TXP, that message should appear before the actual form)

in the tag trace I see this:
```<txp:if_comments_error>
   4022 |      4 |    0.49 |								[<txp:if_comments_error>: false]
```

vanmelick:
Which browser do you use to reproduce this? I can reproduce it only in Dillo, because all my other browsers are too modern (they correctly apply the required attribute on the client-side). I'm having trouble understanding the comment_evaluation class.

phiw13:
Safari 9 - OS 10.11, MobileSafari, iOS 9.2 – both don’t enforce form validation unless you script it.

For Firefox, you can prevent that automatic validation with the `novalidate` attribute on the `form` element.
https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Forms_in_HTML#Constraint_Validation

vanmelick:
The message does appear if the (if_)comments_error tags are placed below the form. It's because the messages are fed to the evaluator during the parsing of the comment input tags, so they're not known before. I'll try to find a fix for this.

vanmelick:
Philippe, can you test pull request #627? I believe it fixes this issue.

phiw13:
Does what it says on the box. Ship it! (and thank you)

IOW - that works fine, both with the default comment_form and with my own variation on it.

philwareham:
Thanks for the report, and fix! Merged.

PR #627, opened 2016-02-14
Improve comment error handling

vanmelick:
Put all the cleanup of comment form field values into the getComment() function. The advantage being that we get the same type of cleanup everwhere we use these values.

Evaluate required fields during preview at a much earlier stage, allowing errors to be shown even if the errors are to be displayed before the comment form fields. Fixes issue #623.

Add language string: comment_duplicate, which is an error message indicating that the comment was rejected, because the message was a duplicate (already exists in database).


philwareham:
Thanks Ruud!

PR #628, opened 2016-02-15
Fix regression when no Expires date is set

vanmelick:


vanmelick:
Wow... forget this one. Looks like I accidentally included a truckload of other changes.

PR #635, opened 2016-02-22
Multi site fix

jools-r:
Edits to README.txt as requested, chunked into different commits for the different new infos/revisions.

Technically an additional note should be made to the readme in txp's base directory about not installing a multi-site installation in a web-readable directory, but I fear that will only cause confusion. Ideas?


philwareham:
Great stuff, thanks. I will merge this branch into master now.

PR #638, opened 2016-02-25
Update multisite index.php to match base index.php

jools-r:
- Reflects change to trace routines for debugging added in ruud's pull update #549
- Maintains path differences for multi-site installation


philwareham:
I'm trusting this is correct, as it matches the main index.php file and @jools-r is experienced with multi-site setups. Merging, thanks.

PR #645, opened 2016-03-07
Fix link to next/prev behaviour

bloatware:
When articles are sorted by some non-unique value (e.g. Section), `<txp:link_to_next/prev />` can produce unexpected results. Same problem admin-side for articles with identical `Posted` value (e.g. cloned). This should fix it in some way.


philwareham:
Thanks Oleg!

Issue #420, opened 2014-10-04
Attribute values of 0 don't render attribute


ghost:
_From [stefdawson](https://code.google.com/u/stefdawson/) on July 01, 2014 12:51:31_

See http://forum.textpattern.com/viewtopic.php?id=41488 Presumably join_atts() is the culprit and:

  $flags & TEXTPATTERN_ATTS_STRIP_EMPTY && !$value

should be more vigorous in its use of 'not a value'. Also affects public tags trying to use attribute values of 0. Will investigate.

_Original issue: http://code.google.com/p/textpattern/issues/detail?id=416_

Bloke:
Is the backwards-compatibility issue big enough to warrant sidelining this, or can we just go ahead and fix it and be damned with people who expect their hand-crafted attributes with a zero value to disappear?

bloatware:
The function `join_atts()` being new, my plan would be to set the default `$flag = TEXTPATTERN_STRIP_EMPTY_STRING` and pass the explicit `TEXTPATTERN_STRIP_EMPTY` in core when necessary. This should suffice for bw compatibility and preserve `attr="0"` in newly created tags by default. The main concern is `tag()` function that is used just everywhere. It makes about 100 places to check :/

bloatware:
A [pull request](https://github.com/textpattern/textpattern/pull/649) is ready. Test, anyone?

Bloke:
Will test tonight. Thanks.

bloatware:
So far so good in my tests. Should we close the issue? Speak now or forever hold your peace!

bloatware:
Should be solved by [this](https://github.com/textpattern/textpattern/issues/649), closing.

PR #649, opened 2016-03-23
Preserve attr="0" in html tags (newly) created with "tag" functions by default

bloatware:
If things go west, throw milestones and revert `function join_atts($atts, $flags = TEXTPATTERN_STRIP_EMPTY_STRING)` back to `function join_atts($atts, $flags = TEXTPATTERN_STRIP_EMPTY)` in `txplib_misc.php`


philwareham:
Tagging against [issue #420](https://github.com/textpattern/textpattern/issues/420)

PR #650, opened 2016-03-24
Adds a class attribute to comments form elements

NicolasGraph:
It's often hard to easily customize the comments form in Txp. There is probably more work to do, but adding a class attribute to the comments form tags could be a beginning…

### DIFF
--- a/textpattern/lib/txplib_forms.php
+++ b/textpattern/lib/txplib_forms.php
@@ -356,6 +356,7 @@ function tInput()
  *
  * @param  string $name     The field
  * @param  string $value    The value
+ * @param  string $class       The HTML class
  * @param  bool   $checked  If TRUE the box is checked
  * @param  int    $tabindex The HTML tabindex
  * @param  string $id       The HTML id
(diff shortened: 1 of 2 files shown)


bloatware:
@NicolasGraph I agree that could be handy, but please consider that

- you should not change function signatures (by inserting a new parameter) of core functions (`checkbox()` here) - that could break backwards compatibility, especially for plugins. You can append `$class` at the end, though, with some default value.
- it's quite possible that these common attributes (`class`, `wraptag` and so on) will be made [automatically available](https://github.com/textpattern/textpattern/pull/484#issuecomment-131627940) for all tags in future versions.

philwareham:
Hi @NicolasGraph, yes as Oleg says this PR cannot be merged in its current form. Any new parameters need to be added at the end - but we are trying to avoid new parameters since they are getting ridiculously long in some cases.

The changes within taghandlers.php are fine, but I would rather see what option the parser optimisation work gives us.

Thanks, and please be patient. I'm sure this functionality can be included at some stage.

NicolasGraph:
I was not sure about the the checkbox function, thank you Oleg and Phil for warn me about that. The new parser could be great; i'll wait it and close this request untill that.

PR #651, opened 2016-03-24
Master sync

bloatware:


vanmelick:
This patch duplicates code unnecessarily, I think. There's no point in having Trace know when to stay quiet if Trace can't act upon it by itself. How much time are we losing on trace calls in PHP 7?

bloatware:
I agree re code duplication, but in PHP 5.4.8 I have measured ~15% long-run difference. We could remove it once PHP 7 is widespread. I just wouldn't like 4.6 to be slower than 4.5.7

vanmelick:
Those that really care about performance, will be using PHP 7. No point in testing on older versions.
Is that a 15% difference comparing just between code fragments or 15% in total runtime? And if the latter, is that on a page with a 'normal' amount of tags or long looping tag constructs? :)

bloatware:
Long loops, for sure! :) It's [here](http://www.iut-fbleau.fr/projet/test/textpattern-parser/articles/) (do few tests). Look at the waiting time in ff network console, I get ~540ms vs 630ms before (your parser branch). And [here](http://www.iut-fbleau.fr/projet/test/textpattern-master/articles/) is the master.

vanmelick:
I don't have time to test this on PHP 7 at the moment and I'm hoping the other devs will comment on this (either way), but if TXP takes the OO-path, then do it properly. Because otherwise, we can probably gain some speed by getting rid of all the other OO stuff as well to get the 4.5.x speed back.

bloatware:
I've also heard that those who really care about performance, will not be using PHP. :) Would be great to see a PHP 7 test, but txp minimal requirements are 5.3 atm, and I see the difference there. It would be pity if the first user impression of 4.6 were "fatter and slower".

Actually, it would be great to gain some speed, but why drop OO? The optimization I suggest has little to do with OO, it's "don't call a function if you know it will not reply". I would suggest the same with the former `trace_add()`. It's not dirty, just looks ugly, until we remove `$quiet` from Trace.

Bloke:
Any time we can not make a function call is okay with me! As you say, a complete waste of resources.

philwareham:
As a non-programmer, I'd say as long as the code is well documented and as lean and easy to maintain as possible, then do what you think is best. The OOP route was pushed hard by Jukka at the time but introduces performance penalties, as he himself acknowledged. If you think the penalties outway the benefits of modularised core then feel free to refactor.

bloatware:
Jukka has done a brilliant work, and I'm in no way into old/new school battle. I just think that letting Trace (be it class or whatever) decide whether it should obey when we say `$trace->log()` is a misconception, including performance-wise. Inherited from the old (though very practical) `trace_add()` misconception.

vanmelick:
Some benchmarks.

``` php
require 'class.trace.php';

$loops = 1000000;
$productionstatus = 'live';
$trace = new Trace();
$trace->setQuiet(true);
$old = microtime(true);

foreach(array('doNothing', 'procedural', 'object') as $function) {
    for ($i = 0; $i < $loops; $i++) $function();
    $new = microtime(true);
    echo sprintf("%10s: %.3f\n", $function, ($new - $old));
    $old = $new;
}

function noTrace() {
    global $trace, $productionstatus;
}

function procedural() {
    global $trace, $productionstatus;
    if ($productionstatus != 'live') {
        $trace->log('something');
    }
}

function object() {
    global $trace;
    $trace->log('something');
}
```

PHP 5.6:
```
   noTrace: 0.349
procedural: 0.393
    object: 0.627
```

PHP7.0:
```
   doNothing: 0.192
procedural: 0.184
    object: 0.242
```

So for PHP7.0 it saves 0.06 microseconds for each place where trace is used. 

To put it in perspective, when Trace is not set to quiet, it takes 26 or 10 seconds for PHP 5 and 7 respectively to run this benchmark.

bloatware:
Impressive, thanks! It saves 0.3 microseconds on PHP 5.4, but this is not enough to explain the 0.1s gain I've observed. Sorry for being a burden, will investigate and try to revert this commits.

bloatware:
The bottleneck seems to be the new Timezone class, or the usage we do of it. These [master](http://www.iut-fbleau.fr/projet/test/textpattern-master/articles/) and [4.5.7](http://www.iut-fbleau.fr/projet/test/textpattern-4.5.7/articles/) test pages now only contain three `<txp:posted />` tags with some tz-sensible formats. I get respectively 650ms vs 515ms processing times. Now, append `?gmt=1` to urls to stop tz_offset calculations, and you get ~430ms for both (and 312ms for parser).

I'm not sure, but it looks like `Timezone->isDst()` is called every time and creates new objects on each call. If so, I think this should be patched before 4.6 is released.

PR #657, opened 2016-04-01
Make pluggable_ui() chainable

bloatware:
This will make plugins (mainly admin-side UI) collaborate rather than compete.


philwareham:
I've [opened a branch](https://github.com/textpattern/textpattern/tree/pr/657) for this, with formatting corrections added.

bloatware:
I find this behavior more natural by default (and flexible enough), but if we fancy 100% backwards compatibility, `pluggable_ui()` can call `callback_event()` twice: once as before, and once new way. Plugins that want to merge will then have to `register_callback()` with `$pre = 1`.

bloatware:
This branch has accidentally (sorry!) accumulated parser commits that I'd like to release for testing. If nobody minds, I will merge it.

philwareham:
Fine by me

Bloke:
Suits me.

Issue #372, opened 2014-10-04
Merge save_pane_state functions into one


ghost:
_From [jukka.m.svahn](https://code.google.com/u/jukka.m.svahn/) on February 24, 2013 10:54:40_

Instead of having those separate functions for each panel, we could calculate and use tokens. The toggle links are generated in PHP, so they could contain fraud preventing token of their own.

By using a token, you will no longer need back-offices, but just that one. This same code could then be re-used for any other async event apart from the super-advanced editors. Like for instance pageby, visible columns etc. could be synched the same way, having no PHP paths of their own.

_Original issue: http://code.google.com/p/textpattern/issues/detail?id=368_

bloatware:
More I look at this, more I think that UI states management should be delegated to client JS (via `localStorage`): no more AJAX requests on every click, nor weird security tokens. Yes, it would be per user+browser rather than per user, but few clicks when you change the browser is not a great annoyance, and there is nothing really secret in panes states. It could even be convenient to store different settings on a widescreen laptop and a phone. Any thoughts?

philwareham:
Sounds logical - if it speeds of/simplifies the code then I'd say do it. localStorage is designed for this sort of stuff anyway.

bloatware:
I have created the [local-storage](https://github.com/textpattern/textpattern/tree/local-storage) branch to test the `localStorage` approach. Seems to work, though I wasn't able to find a `localStorage`-incapable browser for tests.

It eliminates the need for db-storage of panes states in both core and complying plugins. We can test it and close this issue by merely deleting `txp_pane.php` if everything is ok. Meanwhile, I don't see a big risk in keeping it as is.

philwareham:
OK, thanks Oleg. [All modern browsers support `localStorage`](http://caniuse.com/#search=localstorage) so that's great (Opera Mini we don't support anyway, so don't worry about that one).

bloatware:
Thanks Phil, that means testing will be pertinent. It actually should work side by side with the current db storage, so I would merge it in 4.6beta and, following the feedback, drop one of two methods in 4.7.

philwareham:
OK, sounds fine to me.

Please merge if you feel it is OK, but please also create an issue, targeting 4.7, to remind everyone that some of these methods can be removed in the future. Just for reference.

bloatware:
OK, the local storage works so far so good for me. To keep traces of modifications, I defer this issue to 4.7, no particular security threat being involved.

[...]

PR #660, opened 2016-04-07
Whitelist stateful panes

bloatware:
This is intended to close [372](https://github.com/textpattern/textpattern/issues/372), at least concerning whitelisting.


bloatware:
This should prevent unlimited `pane_` prefs population via CSRF by automatically whitelisting all "legal" core panes. If nobody objects, I will close [372](https://github.com/textpattern/textpattern/issues/372) and defer global security remodeling.

bloatware:
It's more complicated than I thought, needs further admin-side homogenization, suspending.

PR #664, opened 2016-04-11
Fix section name descriptions

petecooper:
Section descriptions were numbers (misleading/incorrect), now they’re blank.


philwareham:
I wondered where those numbers came from. Thanks, will merge now.

Bloke:
Good catch. Those values were a throwback to when we had the `is_default` column. Embarrassing I missed it when I added the description column, thanks for the eagle eyes.

PR #681, opened 2016-04-28
Added missing .gitattributes file and updated .gitignore 

natenolting:
Added missing .gitattributes file and updated .gitignore so that PHPStorm .idea folder and composer vendor directory is ignored.


paulbunyannet:
Also added a vagrant box config for testing and development

natenolting:
Also added a vagrant box configuration for testing and development

rwetzlmayr:
I don't think that we should add another 2,903 files to the stock distribution package.

natenolting:
Should I just include the config.yaml file? That can be used to create the vagrant configuration @ puphpet.com, which is where all those files cam from. I usually store whole puphpet folder in the repo so that anyone could grab it and fire up a virtual machine with @vagrant up@, but yea that adds a lot of files.

PR #682, opened 2016-04-29
Save array prefs

bloatware:
See [here](http://forum.textpattern.com/viewtopic.php?id=46969)


textpattern/include/txp_prefs.php
```diff
@@ -120,6 +120,10 @@ function prefs_save()
         if (!isset($post[$name]) || !has_privs('prefs.'.$event)) {
             continue;
         }
+        
+        if (is_array($post[$name])) {
+            $post[$name] = implode(',', $post[$name]);
```
rwetzlmayr: Is it safe to assume that a preference value will never contain a comma?

textpattern/lib/txplib_forms.php
```diff
@@ -134,10 +134,16 @@ function selectInput($name = '', $array = array(), $value = '', $blank_first = f
     $out = array();
 
     $selected = false;
-    $value = (string) $value;
+    $multiple = is_array($value) ? ' multiple="multiple"' : '';
+
```
rwetzlmayr: DocBlock line `@param  string $value` needs fix.

textpattern/include/txp_prefs.php
bloatware: Safe - I think yes, because its items are only compared against `$array` keys, which are under control. A little restrictive, sure, but many txp tokens are comma-separated lists. JSON would be a safer option.

textpattern/lib/txplib_forms.php
bloatware: [Done](https://github.com/textpattern/textpattern/pull/682/commits/e6b344d0a4ba2ee5e776a517b6b4ae2f44b04496), thanks.

bloatware:
@Bloke be my guest :) I have lazily imported `doWrap()` attributes, but totally agree on args unwieldiness. This branch is mainly for functionality testing atm, please feel free to do what you deem necessary with it.

PR #705, opened 2016-06-04
Less fatalist language in default article

petecooper:
Scrub 'fails' -- this stuff is actually stable, after all.


philwareham:
Cheers!

PR #722, opened 2016-07-07
Allow AND/OR in WHERE clause of safe_upsert (+ useless spaces deleted)

NicolasGraph:
As said [here](http://forum.textpattern.com/viewtopic.php?id=47170), the following code does not work.

    safe_upsert(
        'txp_lang', 
        'data = "My data", event = "public"',
        'name = "my_data" AND lang = "en-gb"'
    )

The problem is that on `INSERT`, the `WHERE` clause is simply added to the set clause and throw that:

`[SQL: INSERT INTO txp_lang SET name = 'my_data' AND lang = 'en-gb', data = 'My data', event = 'public' ]`

…It keeps the `AND.`

I know that all `WHERE` clauses won't work with `safe_upsert` but can't we allow the use of `AND` (and `OR`?) like so (or in another better way). `OR` could work if the two columns are different.

### DIFF
--- a/textpattern/lib/txplib_db.php
+++ b/textpattern/lib/txplib_db.php
@@ -394,9 +394,9 @@ function safe_query($q = '', $debug = false, $unbuf = false)
     if ($production_status !== 'live') {
         $trace->start("[SQL: $q ]", true);
     }
-    
+
     $result = mysqli_query($DB->link, $q, $method);
-    
+
     if ($production_status !== 'live') {
         $trace->stop();
     }
@@ -509,6 +509,7 @@ function safe_upsert($table, $set, $where, $debug = false)
     if ($r and (mysqli_affected_rows($DB->link) or safe_count($table, $where, $debug))) {
         return $r;
     } else {
+        $where = implode(', ', (preg_split( "/ (AND|OR) /", $where)));
         return safe_insert($table, join(', ', array($where, $set)), $debug);
     }
 }
@@ -1452,7 +1453,7 @@ eod;
  * the txp_file table.
  *
  * @param  string $type   Column name, lower case (one of 'posted', 'expires', 'created')
- * @param  bool   $update Force update 
+ * @param  bool   $update Force update
  * @return string SQL query string partial
  */
 
@@ -1478,7 +1479,7 @@ function now($type, $update = false) {
             $table = ($type === 'created') ? 'txp_file' : 'textpattern';
             $where = '1=1 having utime > '.$time.' order by utime asc limit 1';
             $now = safe_field('unix_timestamp('.$type.') as utime', $table, $where);
-            $now = ($now === false) ? 2147483647 : intval($now) - 1; 
+            $now = ($now === false) ? 2147483647 : intval($now) - 1;
             update_pref($pref, $now);
             $nows[$type] = $now;
         }


bloatware:
I sometimes do this in plugins, where the queries are under control. But for the core usage the regex pattern looks a bit fragile:
```
safe_upsert(
    'txp_lang', 
    'data = "My data", event = "public"',
    'data = "TWIST AND SHOUT"'
)
```
If you find a more robust pattern, I'll be all for it.

NicolasGraph:
How about [that](http://regexr.com/3dovk)?

bloatware:
That's better, but still will not work with `a=1 AND b=2`. And if we add `\d` (digital) to `('|")` collection as `before AND` delimiter, `c="1 AND 1"` will break it. Tough case...

bloatware:
An easier way would be to pass an array of clauses as `$where` argument.

NicolasGraph:
Yes, why not; otherwise `\d` [seems to work](http://regexr.com/3dovq).

NicolasGraph:
Sorry it's [here](http://regexr.com/3dp00).

bloatware:
Still broken for `a='1 AND one'`. There are also `a=NOW() AND b=1` and the likes to treat.

NicolasGraph:
[This](http://regexr.com/3dp14) is funny to do but I'm not sure if it makes sense to continue…
Any idea how it could work with an array?
Anyway, I can customize `safe_upsert()` in my plugin on my side.

bloatware:
You'd call it like this:
```
safe_upsert(
    'txp_lang', 
    'data = "My data", event = "public"',
    array('name = "my_data"', 'lang = "en-gb"')
)
```
If this is ok, I can submit a (backwards compatible) patch.

vanmelick:
Perhaps even this, so it's not even possible to use any operator other than '=' in the where part:

```
safe_upsert(
    'txp_lang', 
    'data = "My data", event = "public"',
    array('name' => 'my_data', 'lang' => 'en-gb')
)
```

bloatware:
Perhaps, but since we need to insure the bwc, it would be inconsistent with former `name = "my_data"` clauses.

NicolasGraph:
How about the operator? It would always be `AND`?

bloatware:
Only `AND`, I guess, unless we agree on some natural logic to assign values when `a=1 OR b=2` is not found.

NicolasGraph:
For me, this _kind_ of thing (customized for clarity):

    safe_upsert(
        'txp_lang', 
        'data = "My data", event = "public"',
        'name = "my_data" OR lang = "en-gb"'
    )

would assign _my_data_ to the column _name_ and _en-gb_ to _lang_.
However this could be a problem:

    safe_upsert(
        'txp_lang', 
        'data = "My data", event = "public"',
        'name = "my_data" OR name = "your_data"'
    )

…?

philwareham:
Bump? Is this targeted for 4.6?

bloatware:
> Is this targeted for 4.6?

Not really, I think. It would make plugins authors life a little more convenient, but we (well, I) don't see a natural way to satisfy all the requirement. The best I can do now is follow Ruud's proposal and treat only `AND` operator, if nobody finds a better idea.

NicolasGraph:
I agree, let see if we can find something better for 4.7, pending that plugin authors can build a custom function. I can close this pull request and open an issue…?

NicolasGraph:
[To be continued…](https://github.com/textpattern/textpattern/issues/724)

jools-r:
I just tripped over this one too (also in combination with setting the language). Has there been a solution to this, or does one have to do this in two stages:
- check if it exists: 
  - if yes, do safe_update with where A AND B, 
  - if no, safe_insert with all the columns.

bloatware:
It should work with `AND` arrays as [suggested](https://github.com/textpattern/textpattern/pull/722#issuecomment-231193887) by Ruud.

jools-r:
Thanks Oleg. That did it!

PR #734, opened 2016-07-29
Disable email address publishing in Atom by default

petecooper:


petecooper:
Merged before I could even update OP. Blimey.

Typing is hard! Thanks, @Bloke!

PR #735, opened 2016-07-29
Tidying HISTORY.TXT

petecooper:
Formatting and arrangement, plus change relating to Atom feeds in https://github.com/textpattern/textpattern/pull/734

### DIFF
--- a/HISTORY.txt
+++ b/HISTORY.txt
@@ -1,6 +1,6 @@
 Changes in 4.6.0 (Upcoming)
 
-* Textile v3.5. Also support for adding other text filters via plugins (for example, Markdown).
+* Textile v3.5.
 * Improved RTL language support.
 * Improved admin side UI.
 * Requires PHP 5.3.3 or newer. Compatible with PHP 7 and MySQL 5.7 (thanks Ruud van Melick).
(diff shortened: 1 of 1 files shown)


HISTORY.txt
```diff
@@ -9,45 +9,45 @@ Changes in 4.6.0 (Upcoming)
...
+* Changed: <txp:comment_submit /> accepts 'label' attribute.
+* Changed: <txp:comment_web_input /> accepts 'size' attribute.
 * Changed: <txp:link /> accepts 'id' and 'name' attributes.
 * Changed: <txp:meta_author /> accepts 'escape' and 'format' attributes.
 * Changed: <txp:meta_keywords /> accepts 'escape', 'format' and 'separator' attributes.
-* Changed: <txp:password_protect /> tag can now be used as a container tag.
+* Changed: <txp:password_protect /> can be used as a container tag.
 * Changed: <txp:recent_articles /> accepts 'offset' attribute.
 * Changed: <txp:section_list /> accepts 'html_id', 'limit' and 'offset' attributes.
-* Changed: Deprecated <txp:keywords /> tag. Use <txp:meta_keywords format="" /> instead.
```
Bloke: This line was a shallow attempt at documentation, because the direct analog of the old `<txp:keywords />` tag is `<txp:meta_keywords format="" />`. Without the attribute, you get different output.

Your edit is neater, because such usage should arguably be in the docs. But we have (perhaps wrongly) put documentation in the HISTORY file before: see the Changes in 4.0.6 section where this kind of thing was written:

    Deprecated tag: <txp:s /> replaced by <txp:page_url type="s" />.

I don' t know if anyone reads this HISTORY prior to upgrade, but on the offchance they do, such hints may be worth keeping as a quick reference so upgraders know they may have to make changes to their site tags.

Don't really mind either way. If docs should be confined to the docs, I'm cool with that.

petecooper: Agreed. Reverted.

PR #738, opened 2016-07-29
Switch hardcoded diag text for gTxt

petecooper:
Will add Textpack translation if merged.


petecooper:
Is messy. Will close and refactor.

Issue #412, opened 2014-10-04
Implement PRS-0/PSR-1/PSR-2 and automatic styling check


ghost:
_From [jukka.m.svahn](https://code.google.com/u/jukka.m.svahn/) on April 05, 2014 17:12:49_

We should use the accepted standard rather than doing our own thing.

* https://github.com/php-fig/fig-standards/tree/master/accepted

_Original issue: http://code.google.com/p/textpattern/issues/detail?id=408_

Bloke:
Closing this now as we've adopted PSR-2 throughout. Any latent issues will be picked up by PHPCS or contributors during development.

PHPDoc blocks still need adding in places (see #278), but doc formatting is not part of PSR-2.

NicolasGraph:
I started to use [linter-phpcs](https://atom.io/packages/linter-phpcs) and it seems that PSR-2 is not totally adopted by Txp yet…
[Here](https://github.com/NicolasGraph/textpattern/commit/5c13ee2b84ad83cc8bcd151c5b46d113617121f1) are needed changes just for _txp_admin.php_.
Now, let me know: I missed something or you need help for cleaning?

philwareham:
@NicolasGraph PSR-2 is done as far as it needs to be at this stage, it's an ongoing task - personally I've used [php-cs-fixer](http://cs.sensiolabs.org) and that throws up quite a lot of changes - I just cherrypicked the ones that were most appropriate/important.

NicolasGraph:
Ok, just let me know if you need someone to help in that easy boring thing to do in a near future.
See you!

NicolasGraph:
…Sorry, I can't get this question out of my mind: what does it mean _PSR-2 is done as far as it needs to be at this stage_? It is a standard, you follow it or not. no?

Bloke:
The trouble with automatic linters is they're not very smart. Last time I ran one, it ballsed up my admin side because it put line breaks where it shouldn't have (in JSON output), which introduced syntax errors and bugs.

We've been taking a softly-softly approach and applying PSR-2 where it makes sense. But in some places, e.g. separating every parameter onto its own line, it makes the code far too spread out and harder to comprehend.

Like all good standards, there are things we try to follow to the letter of the law (e.g. bracket / brace placement, spacing, indentation, etc) and other places where we employ a little wiggle room (e.g. not always putting each parameter of a function call on its own line).

[...]

PR #740, opened 2016-07-30
PSR-2 coding style

NicolasGraph:
Related to [this](https://github.com/textpattern/textpattern/issues/412).

It's a first step and It can still be impoved; I didn't change constant, class and method name letter cases for example, and some changes can be discussed but I think that it is a bit better now.

Some things I'm thinking and that could be discussed 
- a lot of `td()` are now inline but you could want to separate these lines (see [here](https://github.com/NicolasGraph/textpattern/blob/master/textpattern/include/txp_file.php#L436));
- In most cases I kept `column_head()` arguments on two lines because the last on is the only one on the second line and others are shorts, so I thought you could want to keep it as is (see [here](https://github.com/NicolasGraph/textpattern/blob/master/textpattern/include/txp_file.php#L319);
- each last and short form() arguments are on one line, maybe I should keep them together (see [here](https://github.com/NicolasGraph/textpattern/blob/master/textpattern/include/txp_file.php#L228).

### DIFF
--- a/css.php
+++ b/css.php
@@ -40,7 +40,8 @@ if (@ini_get('register_globals')) {
         (array) $_POST,
         (array) $_COOKIE,
         (array) $_FILES,
-        (array) $_SERVER);
+        (array) $_SERVER
+    );
 
     // As the deliberate awkwardly-named local variable $_txpfoo MUST NOT be
     // unset to avoid notices further down, we must remove any potential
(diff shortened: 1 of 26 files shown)


NicolasGraph:
F…, it seems there was a bug on few files; I need to check, sorry.

PR #750, opened 2016-08-31
Removing deprecated warning trace_add()

makss:
Related topic: http://forum.textpattern.com/viewtopic.php?id=46841


philwareham:
@bloke @bloatware is this OK to merge? Want to release beta3 today.

bloatware:
I would not (yet) deprecate `trace_add()`, until the new Trace gets a final shape. So OK to merge for me, but only temporary.

philwareham:
OK, I think the better route would be to put a comment on this line that it is marked for deprecation, not remove it during this release. closing but thanks anyway.

PR #754, opened 2016-09-04
`README.txt` tweaks

petecooper:
OS X will become macOS soon, plus uniformity tweaks.


philwareham:
Yep. MacOS Sierra is imminent (and already on public beta).

Issue #755, opened 2016-09-04
References to textpattern.net


petecooper:
I'm doing a docs shakedown. There's a reference to textpattern.net (wiki) - is it safe to link to http://docs.textpattern.io instead, yet?

philwareham:
It's safe, if a little messy and unfinished. I'm hoping work gets a little quieter in next couple of weeks so I can spare a few hours to get it tidied. It's still better than the wiki though so please link to new docs site from here on.

philwareham:
Still some to do :) I'll amend and then close again.

philwareham:
Done now. Closing.

PR #757, opened 2016-09-05
Switch another docs ref away from wiki to .io

petecooper:
See https://github.com/textpattern/textpattern/issues/755 for info.


philwareham:
Thanks Pete.

Issue #775, opened 2016-09-26
[BUG - 4.6.0] Setup: No preferences sql_now_*


makss:
After the setup does not create preferences in `txp_prefs` table:
* sql_now_posted
* sql_now_expires
* sql_now_created

This bug causes the update attempt `txp_prefs` table for each access to the site.

PR #776, opened 2016-09-26
Bugfix: No preferences sql_now_*

makss:
Fix Issue: #775 
This patch should be applied to `dev` branch, too.


Bloke:
I thought these prefs were soft-created on first call of the function? Does it cause problems if they don't exist already?

makss:
These preferences are only updated. For faster operation did not make additional checks.
For updates already have code in `/textpattern/update/_to_4.6.0.php`

```
// Add prefs to allow query caching when now() is used.
if (!get_pref('sql_now_posted')) {
    set_pref('sql_now_posted', time(), 'publish', PREF_HIDDEN);
    set_pref('sql_now_expires', time(), 'publish', PREF_HIDDEN);
    set_pref('sql_now_created', time(), 'publish', PREF_HIDDEN);
}

```

In other issue I propose AutoFix similar cases.

bloatware:
> These preferences are only updated.

Yes, and it does not feel right. We should create these preferences on setup, but if later they are deleted (accidentally, say), the problem will rise again. Rather than relaying on Diagnostics tab visit, I would suggest that important preferences are auto-repaired when needed.

PR #782, opened 2016-09-27
A more complete SQL tracing

makss:
* Record the number of rows returned by the SQL query.
* Remove unused `global $txpcfg`


bloatware:
Good idea. Maybe simply
```
$trace->stop("[Rows: ".intval(@mysqli_num_rows($result))."]");
```

makss:
@bloatware , You're right. ;-)

Issue #774, opened 2016-09-24
[BUG - 4.6.0] Cannot change Author on Articles


Stoyvo:
I have the following setup:
User 1: Publisher
User 2: Staff Writer
User 3: Staff Writer

Currently all my articles are marked with the Author as User 1, even if User 2 and User 3 have created articles.

In Content -> Articles, using the Change Author action does not apply the selected author.

There are no plugins installed, never has there been. Our initial blog was created with 4.6.0-beta, I'm only creating the ticket as I see it's still in effect on 4.6.0 stable release.

Bloke:
Sorry, I cannot reproduce this. Would you be able to supply some more detail about your environment please? PHP / MySQL version, and any other pertinent info that might help us track down if this is a general bug or specific to a particular hosting setup. Thanks.

Stoyvo:
PHP Version: PHP Version 7.0.10
Database: MariaDB 5.5.52

makss:
I got to reproduce this situation.
Now `Staff Writer` (priv = 4) can not change `AuthorID` even on their articles.
```
    'article.edit'                => '1,2,3',
    'article.edit.published'      => '1,2,3',
    'article.edit.own'            => '1,2,3,4,5,6',
    'article.edit.own.published'  => '1,2,3,4',
```

Code: https://github.com/textpattern/textpattern/blob/dev/textpattern/include/txp_list.php#L637
```
        // Change author.
        case 'changeauthor':
            $value = ps('AuthorID');
            if (has_privs('article.edit') && in_array($value, $all_authors, true)) {
                $field = 'AuthorID';
            }
            break;
```

There are two options:
* Hide `changeauthor` option for `Staff Writer` from drop down list.
code: https://github.com/textpattern/textpattern/blob/dev/textpattern/include/txp_list.php#L566
```
    if (has_single_author('textpattern', 'AuthorID')) {
        unset($methods['changeauthor']);
    }
```
change to
```
    if (has_single_author('textpattern', 'AuthorID') || !has_privs('article.edit')) {
        unset($methods['changeauthor']);
    }
```

* Or allow `Staff Writer` change `AuthorID` on their articles.
```
        // Change author.
        case 'changeauthor':
            $value = ps('AuthorID');
            if (has_privs('article.edit.own.published') && in_array($value, $all_authors, true)) {
                $field = 'AuthorID';
            }
            break;
```
and
```
[...]

PR #783, opened 2016-09-28
Show option "Change author" only if the user has rights

makss:
Issue: #774 

A similar situation may occur with: `image`, `link`, `file`. Now there are not checked right to change the author.
Probably it is necessary add a check of the rights: `image.edit`, `link.edit`, `file.edit` ?


Bloke:
Yes, I guess it should be the same for the other three panels. Do you fancy patching those too for us, or shall I do it?

PR #792, opened 2016-10-03
Enable pageby and pgonly for <txp:article_custom /> tag

bloatware:
With pgonly="1" <txp:article_custom /> will output the number of matching articles pages, without setting the pagination.

### DIFF
--- a/textpattern/publish.php
+++ b/textpattern/publish.php
@@ -727,8 +727,6 @@ function doArticles($atts, $iscustom, $thing = null)
             'searchform'   => '',
             'searchall'    => 1,
             'searchsticky' => 0,
-            'pageby'       => '',
-            'pgonly'       => 0,
         );
     }
 
@@ -746,6 +744,8 @@ function doArticles($atts, $iscustom, $thing = null)
         'frontpage'     => !$iscustom,
         'match'         => 'Category1,Category2',
         'offset'        => 0,
+        'pageby'        => '',
+        'pgonly'        => 0,
         'wraptag'       => '',
         'break'         => '',
         'label'         => '',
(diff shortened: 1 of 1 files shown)


makss:
It seems already been attempts to add `pgonly` attribute in `article_custom` tag.

bloatware:
> It seems already been attempts to add pgonly attribute to article_custom tag.

Sure, including mine, few years ago :-) It didn't get much attention, but I'm not desperate. The aim is to be able to easily count pages or articles (with `pageby="1"`) for informational purposes or for use in pagination plugins.

bloatware:
The initial patch (enable pages count for article_custom) is merged into dev now. The rest of this branch allows for finer custom fields match in article(_custom) and related_articles.

Bloke:
Although I haven't tested it, this seems like a pretty good idea to me. Is it okay to be merged? Any backwards-compatibility issues?

Regarding the [item attribute](https://github.com/textpattern/textpattern/pull/792/commits/f3083ccb000660329a0f7d02e5c574357b2d9c13): it assumes comma is the separator. But an [earlier commit](https://github.com/textpattern/textpattern/pull/792/commits/91fd77b1b850865c41c7ce36e9ec3945186d1b05#diff-0b3cef3dffcc8a9774dfa6d412340b4aL4961) references pipe also. Should we introduce a `separator` attribute here, like there is for the `if_custom_field` tag, so that people are free to delimit their custom fields in any way they see fit? Or is that too confusing?

bloatware:
Thanks @Bloke. I thought we should postpone it to the unlimited custom fields era, also because `separator` could then become part of custom field definition.

petecooper:
Paging @bloatware – what's the status of this branch / PR from your viewpoint? The branch reports as 20 commits ahead of `dev`, are those commits worth evaluating for inclusion in `dev`, or would you prefer it be left open?

bloatware (commented):
Thanks @petecooper, will merge after 4.8.8 freeze.

Issue #796, opened 2016-10-08
Hide checksums.txt by default


Bloke:
While there are many ways to determine which version of Txp is running on a server, should we close one simple mechanism: make `checksums.txt` inaccessible to the outside world? All it would involve is:

```
<Files checksums.txt> 
deny from all 
</Files>
```

added to `textpattern/.htaccess`. Diagnostics still runs fine because it reads the file directly. This change just makes one less essential file exposed to prying eyes. Overkill? Useful?

bloatware:
Actually, would allowing just `index.php`, `textpattern.[js|css]` and `admin-themes` folder work?

makss:
Maybe allow only `index.php` and static files?
```
order deny,allow
deny from all

<Files ~ "\.(js|css|svg|png|jpg|gif|ico)$">
        allow from all
</Files>

<Files ~ "^(index\.php|)$">
        allow from all
</Files>
```

Or it is possible use mod_rewrite for this.

rwetzlmayr:
> Actually, would allowing just `index.php`, `textpattern.[js|css]` and `admin-themes` folder work?

No. Consider to add at least `css.php`.

makss:
`css.php`  deprecated in 4.2.0 (2009-08-28) "This file is here for backwards compatibility."

It has been seven years in Textpattern's been a lot of changes. I do not know how this file can help with backward compatibility. Perhaps it is to remove.

rwetzlmayr:
Wrong planet. Look [here](https://github.com/textpattern/textpattern/blob/dev/css.php).

makss:
I mean `textpattern/css.php`.  [link](https://github.com/textpattern/textpattern/blob/dev/textpattern/css.php)
I fixed my post above.

bloatware:
On second thoughts, we don't know what files plugins will need to access. So generic `deny from all` rules may be too restrictive.

Bloke:
> On second thoughts, we don't know what files plugins will need to access. So generic deny from all rules may be too restrictive.

Yeah, it would break MLP for starters. Well, alright, it's broken anyway but you know what I mean...

petecooper:
Putting my pen tester (white) hat on, there are a bunch of trivial ways a Textpattern site's version number can be determined.

[...]

PR #798, opened 2016-10-11
Configure web access to /textpattern folder

makss:
Discussion of the Issue #796 

ToDo: remove file `/textpattern/css.php`

### DIFF
--- a/textpattern/.htaccess
+++ b/textpattern/.htaccess
@@ -6,13 +6,13 @@
     RewriteEngine On
 
     # Uncomment if you need web access to some folder
-#    RewriteCond %{REQUEST_URI} ^/textpattern/some_folder [OR]
+#    RewriteCond %{REQUEST_FILENAME} /textpattern/some_folder [OR]
 
     # Allow web access to /textpattern/index.php and /textpattern/setup/index.php
-    RewriteCond %{REQUEST_URI} ^/textpattern/(setup/)?(index\.php)?$ [OR]
+    RewriteCond %{REQUEST_FILENAME} /textpattern/(setup/)?(index\.php)?$ [OR]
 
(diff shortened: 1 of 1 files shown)


bloatware:
I have tried it, but was denied all admin-side access.

makss:
@bloatware  , please try again.

bloatware:
Thanks, but this still looks too restrictive. In particular, some plugins could break (see [this comment](https://github.com/textpattern/textpattern/issues/796#issuecomment-252899898))

makss:
@Bloke 
> Yeah, it would break MLP for starters.

I have installed Textpattern: 4.7.0-dev, then install [MLP](https://github.com/bloke/mlp/tree/v4.6.0) plugin.
(I had a little to correct Textpattern code to the plugin has started)

MLP plugin correctly makes web requests through `textpattern/index.php`
I have not seen direct web requests to the other php file.

I did not see the problems associated with my pull request. Correct me if I'm wrong.

bloatware:
> I did not see the problems associated with my pull request. Correct me if I'm wrong.

We can not exclude that some plugin (image processing, polls, etc...) will need to initiate a browser request (say, ajax) to external php files. @Bloke could say more about MLP, but I have already met such situations. I'm not sure an average user would be able to adapt `.htaccess` in this case.

bloatware:
`css.php` is removed, thanks. New `.htaccess` would need adding `rpc/index.php` exception, but I would rather close the request. @makss ?

makss:
> New `.htaccess` would need adding `rpc/index.php` exception, but I would rather close the request.

This pull request sets web access only to the folder `/textpattern`, `/rpc` is outside.

Yes, close.

PR #803, opened 2016-10-15
txp_tag: normalises $_POST value

makss:
Normalises MAGIC_QUOTES_GPC and removes NULL bytes for $_POST['linktext']


bloatware:
Yes, this one could go into 4.6.2, imo. Kind of bugfix.

Issue #802, opened 2016-10-15
Admin usability: required attribute for input fields


makss:
It would be convenient to mark the required fields for input.
`<input ... required="required" />`
`<textarea required="required" >` (plugin tab)

And visually highlight them using css:
```
input:required, textarea:required {
        background-color: #...;
// or
        box-shadow: 0 0 5px #... inset;
}
```

Already done:
event=page&step=page_new
event=form&step=form_create
event=css&step=pour

Need to do:
event=section&step=section_edit
event=article
event=link&step=link_edit
event=category
event=admin&step=author_edit
event=plugin
Login form
Reset password form

philwareham:
Any CSS amends need to be made on the theme repos, not core. Open an issue at https://github.com/philwareham/textpattern-hive-admin-theme for that part. Cheers!

philwareham:
I've added some CSS visual feedback for required fields in the admin-side now (dev branch only), any PHP changes would need to be done by @bloke or @bloatware though.

PR #805, opened 2016-10-17
Required attribute for input fields

makss:
Issue #802 

Please do not apply this patch is, I want to check out the other form.


makss:
It seems checked all the forms, the patch can be applied.

Issue #787, opened 2016-09-29
<txp:hide pass="1">


makss:
Simple things that make magic. Why no one has done so far is it?

In the first pass are calculated(set) variables, they are displayed or processed in a second pass. I have long been using a similar approach, but with the help of other tools(wrapper based on `ob_start()`).

I bring in one form of content, title, metas, connect additional js and css. During the second pass - all these elements are in the right places page.
If there is a new content type, I simply add another form. All processing of a new type of content placed in one form, I do not need to remember where and what conditions are set to have the right titles, metas, etc.

The example is not entirely intuitive, but...

```
<head>
<txp:variable name="mytitle" value='<txp:site_name />' /> <!-- set default values -->
...

<txp:hide pass="1">
    <title><txp:variable name="mytitle" /></title>
    ...meta_description
    ...meta_keywords
    ...meta_robots
    ...canonical
    ...meta_author
    ...extra include css
</txp:hide>
</head>
<body>
<txp:if_section>
    output section content
    <txp:variable name="mytitle" value="some specific title" />
    set specific meta values, which are displayed in the `<head>` in the secondpass: title,
description, keywords, etc...
</txp:if_section>

Some of the included form...
<txp:if_search>...output content && set metas...</txp:if_search>
<txp:if_individual_article>...output content && set metas...</txp:if_individual_article>

<txp:hide pass="1">
    <txp:variable name="my_extra_js" />
    ...extra include js
</txp:hide>
</body>
</html>
```

bloatware:
Nice idea, though not quite new :-)

* [adi_if_content](http://forum.textpattern.com/viewtopic.php?id=46104) plugin can alter the processing order;
* we are considering the introduction of universal `txp` tags attributes, one of them being `txp:order`. It will indicate in which order tags are processed. With them you should be able to do things like this (currently impossible):
```
	<txp:newer txp:order="6">Newer</txp:newer>
	<txp:older txp:order="6">Older</txp:older>
	<txp:article /> <!-- with default txp:order="5" -->
```
This said, `<txp:hide />` tweak is fine for me, but I don't know what are our plans re `secondpass` (Jukka was for removing it).

makss:
Yet another way run the tag during the second pass. Same as `adi_if_content insert=""`  (If I get you right.)
```
<txp:variable name="secondpass" value="<txp:title />" />
<txp:variable name="secondpass" />
```


[...]

PR #806, opened 2016-10-17
Adding `pass` attribute to txp:hide tag

makss:
Pull to `dev` branche
Issue: #787

### DIFF
--- a/textpattern/publish/taghandlers.php
+++ b/textpattern/publish/taghandlers.php
@@ -5013,7 +5013,16 @@ function file_download_description($atts)
 
 function hide()
 {
-    return '';
+    global $pretext;
+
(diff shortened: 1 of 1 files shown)


bloatware:
If we modify `<txp:hide />`, I suggest adding some conditionally hiding abilities, à la [adi_if_content](http://forum.textpattern.com/viewtopic.php?id=46104): hide the entire block if some tags inside are "empty".

bloatware:
I'm closing this request, since `<txp:hide />` has [evolved](https://github.com/textpattern/textpattern/commit/f9919e1a0496aa40d1e0c4a3ddad8fba4fab830f), but the `secondpass` idea will be implemented soon for all tags. Thanks @makss!

PR #811, opened 2016-10-23
Swap apache_request_headers() for $_SERVER

Aeyoun:
Apache isn’t the only server in town anymore, and there is no reason to use this method rather than the superglobal variable.

### DIFF
--- a/textpattern/publish/atom.php
+++ b/textpattern/publish/atom.php
@@ -298,14 +298,8 @@ function atom()
         $hims = serverset('HTTP_IF_MODIFIED_SINCE');
         $imsd = ($hims) ? strtotime($hims) : 0;
 
-        if (is_callable('apache_request_headers')) {
-            $headers = apache_request_headers();
-
-            if (isset($headers["A-IM"])) {
-                $canaim = strpos($headers["A-IM"], "feed");
-            } else {
-                $canaim = false;
-            }
+        if (isset($_SERVER["HTTP_A_IM"])) {
+            $canaim = strpos($_SERVER["HTTP_A_IM"], "feed");
         } else {
             $canaim = false;
         }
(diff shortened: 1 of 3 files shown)


philwareham:
@Bloke @bloatware @vanmelick is this OK to merge? Not my area.

Aeyoun:
Reviving this should only take five seconds, if anyone is up for it. The patch simply replaces three uses of the `apache_request_headers()` function  with the `$_SERVER` superglobal.

Bloke:
Honestly I don' t know what A_IM is or does, nor can I find any reference to it on the Internet. But if Aeyoun says it's available in `$_SERVER`, and we need to still use it, then who am I to argue? The superglobal approach is fine by me, since it doesn't appear we're using `apache_request_headers()` for any other special values. But, as I say, I have zero knowledge in what this bit of code does for Txp.

Aeyoun:
Every HTTP request header is in $_SERVER servers. The difference from `apache_request_headers()` is that the superglobal prefixes headers with `HTTP_` and that it is available in every web server environment and not just those claiming to be Apache. This patch just makes sure everyone is using the same variables, regardless of server environment.

Now that I’m looking more closely into the feed delta implementation (`A-IM`) in Textpattern; I see that it’s actually entirely broken. That is a separate issue, however. I can fix that in a few days time, it shouldn’t be much work at all.

Bloke:
Thanks for the clarification. And if you can fix the feed implementation as part of this request then that'd be brilliant. It's the dark arts to me.

Would you prefer we merge this one and then you can make a separate PR for the next fix or do you want to add more commits to this one? I'm happy either way.

Aeyoun:
Just merge this, and I’ll get around to the other thing later this week. 👍

makss:
It is advisable to rewrite the code in `atom.php` and `rss.php`, very much there is out of date and no longer makes sense. Can be deleted:
* completely block `apache_request_headers()`  or $_SERVER["HTTP_A_IM"]
* completely block `$cutarticles` and related headers.  (Internet is not running on the modems and such complexities in order to save traffic, I do not see the point.)
* Remove function safe_hed
* Remove function fixup_for_feed

Simplified calculation Etag 
It seems that there still can simplify sql queries.

Aeyoun:
Pull request #831 takes care of the remainder, as promised.

PR #812, opened 2016-10-23
Uncomment and enable HTTP 226 responses

Aeyoun:
This was a problem with Apache 1.3 (since patched), but Apache 2+ and other web servers don't have a problem with this anymore.

### DIFF
--- a/textpattern/publish/atom.php
+++ b/textpattern/publish/atom.php
@@ -350,8 +350,7 @@ function atom()
         }
 
         if ($cutarticles) {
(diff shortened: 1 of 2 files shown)


philwareham:
I'm OK with this, since Apache 2 has been around for 10+ years now and last Apache 1 release was 9 years ago. @Bloke and @bloatware do you concur?

bloatware:
I guess it's ok.

philwareham:
Thanks again!

Issue #815, opened 2016-10-24
Deprecated code


philwareham:
Since we don't ever have major point releases (such as 5.0) - the standard process of removing deprecated code never really happens. We should (selectively) start getting rid of redundant code that was deprecated in Textpattern 4.5 release and earlier now.

philwareham:
Done the obvious stuff, no doubt there is code that hasn't been documented yet that is also deprecated. If it ever gets documented or flagged I'll delete on case-by-case basis.

PR #816, opened 2016-10-24
Deprecated stuff

philwareham:
See https://github.com/textpattern/textpattern/issues/815 - removed some code that was know to be deprecated  prior to Textpattern 4.6


bloatware:
By any mean, if this is not used anymore, trash it!

PR #817, opened 2016-10-24
Single form of <txp:if_tags />

bloatware:
Returns true or false, output as '1' or '' when parsed.

### DIFF
--- a/textpattern/publish/taghandlers.php
+++ b/textpattern/publish/taghandlers.php
@@ -1888,7 +1888,7 @@ function article_url_title()
 
 // -------------------------------------------------------------
 
-function if_article_id($atts, $thing)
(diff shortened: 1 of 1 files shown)


philwareham:
Can you provide a usage example so I can update the documentation? Cheers.

bloatware:
Sure, thanks, Phil. You might say that `<txp:if_something />` is equivalent to `<txp:if_something>1</txp:if_something>`. Example:
```
<txp:comments_invite textonly='<txp:if_comments_disallowed />' />
```
is the same as
```
<txp:comments_invite textonly='<txp:if_comments_disallowed>1</txp:if_comments_disallowed>' />
```
but shorter and easier to parse.

PR #820, opened 2016-10-25
Tell robots to not index error pages

Aeyoun:
Provide a hint to search engines that errors shouldn’t be index. Will clear once the error is resolved.


philwareham:
Yep, good idea, thanks!

philwareham:
I reverted my revert on this commit since error pages have been indexed in search engines. This meta gives extra protection from that.

PR #821, opened 2016-10-25
Use rel="nofollow" in “Published with” footer message

Aeyoun:
Textpattern is currently violating [Google’s link building scheme](https://support.google.com/webmasters/answer/66356?hl=en) rule:
* Widely distributed links in the footers or templates of various sites

Apply nofollow to link to avoid penalty.

### DIFF
--- a/textpattern/setup/pages/archive.txp
+++ b/textpattern/setup/pages/archive.txp
@@ -178,7 +178,7 @@
         <p>
             <small>
                 <txp:text item="published_with" />
-                <a rel="external" href="http://textpattern.com" title="<txp:text item="go_txp_com" />">Textpattern CMS</a>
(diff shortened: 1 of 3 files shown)


philwareham:
This would actually need to be done in the actual default theme repo, not the core: https://github.com/textpattern/textpattern-default-theme

Thanks! Closing here.

Aeyoun:
Why is the theme them duplicated in this repository? It should be included in this repo as a git subrepository to avoid this very problem. I’ll try again in the correct repository.

philwareham:
Because we were eventually planning on using Composer to include theme packages, but since Textpattern doesn't have a build script right now that hasn't happened yet.

Git submodules pull the whole repository in, when we would only require the `dist`contents, so that would be undesirable here.

PR #822, opened 2016-10-25
Include feed type="" attr in feed_link

Aeyoun:
Helps designers target feed links more easily with CSS, and help robots and software figure out what the link leads to without following it.

All other major CMS provide this at least in their default themes.


bloatware:
Please switch to more 4.6-ish
```
$out = href($txt, $url,  array( 'type' => $type, 'title' => $title ));
```

Aeyoun:
@bloatware, like this?

bloatware:
Fine, thanks.

PR #823, opened 2016-10-26
Redirects should have a linked body message

Aeyoun:
This is dead code, as far as I can tell; but fixing it anyway. Not all clients follow redirects automatically.

Redirect pages generated by all major web servers include a body message for redirects.

Users see informational page instead of blank page when redirect fails because of caching, network issues, tweaked browsers, accessibility settings, misuse of curl, etc. etc.

### DIFF
--- a/textpattern/lib/txplib_misc.php
+++ b/textpattern/lib/txplib_misc.php
@@ -5026,7 +5026,7 @@ function txp_die($msg, $status = '503', $url = '')
     if ($url && in_array($code, array(301, 302, 303, 307))) {
         ob_end_clean();
         header("Location: $url", true, $code);
-        die('<html><head><meta http-equiv="refresh" content="0;URL='.txpspecialchars($url).'"></head><body></body></html>');
+        die('<html><head><meta http-equiv="refresh" content="0;URL='.txpspecialchars($url).'"></head><body><p>Document has <a href="'.txpspecialchars($url).'">moved here</a>.</p></body></html>');
     }
 
     if ($connected && @txpinterface == 'public') {


bloatware:
@Aeyoun good idea, and I would merge it if we had localized `moved_to` strings. Or output just the link?

Aeyoun:
I thought about it, bu decided aginst it. It’s a single lone sentence on an empty document with one link that containes the word “here”. I don’t think there’ll be much of a lingual barrier if anyone stumble across this. I  see “Click here” is used on links even on Chinese language webpages. No one should see this message unless there is a network problem, special needs (accessibility), or someone using curl without following redirects.

I could switch it to just a link, but the text with link is probably a better user experience than just a linked URL — even if you don’t fully understand the language.

philwareham:
This is no different than how the systems states "Database unavailable' only in English.

bloatware:
@philwareham  I leave it with you to decide if it's ok UX-wise, but think that a localized [Permanent link to this article](#) string could be an option. _I_ would see this message, since redirects are disabled in my daily browser.

philwareham:
I can create a new translation string `moved_to` as suggested, how about that?

bloatware:
Nope for me, new strings are rarely translated nowadays.

philwareham:
Well, I don't agree with that, we have a [fairly healthy level of translation activity](https://github.com/textpattern/textpacks/commits/master).

I'm also building a translations page into the new Textpattern official site, which encourages people to use our [Crowdin](https://crowdin.com/project/textpattern-cms-textpacks) project.

bloatware:
Ah, sorry, nice to hear it! But I'd prefer to keep the number of strings as low as possible. Merge it as is, maybe?

philwareham:
Yep, agreed. Done. Thanks everyone!

Aeyoun:
> @bloatware would see this message, since redirects are disabled in my daily browser.

I guess that confirmed that this is dead code we’re discussing. You would have noticed this otherwise. Used by plugins, maybe?

bloatware:
[txp_die()](http://phpcrossref.com/xref/textpattern/_functions/txp_die.html) is used quite often, though not with `3xx` status. Plugins, maybe.

PR #824, opened 2016-10-26
WordPress-style URL alias ?feed=1 and /feed for Atom feeds

da2x:
Allows for guessable URLs for feeds. No redirect as that would waste bandwidth and slow down performance. Content duplication already handled by the  `atom:self` element for Atom and RSS.

`/feed` is a common endpoint for syndication feeds used by *a lot* of content management systems and websites.

### DIFF
--- a/textpattern/publish.php
+++ b/textpattern/publish.php
@@ -231,12 +231,12 @@ function preText($s, $prefs)
     // Set messy variables.
     $out = makeOut('id', 's', 'c', 'context', 'q', 'm', 'pg', 'p', 'month', 'author');
 
-    if (gps('rss')) {
-        $out['feed'] = 'rss';
+    if (gps('atom') || gps('feed')) {
+        $out['feed'] = 'atom';
     }
 
-    if (gps('atom')) {
-        $out['feed'] = 'atom';
+    if (gps('rss')) {
+        $out['feed'] = 'rss';
     }
 
     // Some useful vars for taghandlers, plugins.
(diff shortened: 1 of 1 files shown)


makss:
Why add hard code? Use if you want `.htaccess`.

Textpattern already have enough hard code and actually get it off (for Wordpress this is true, too).

rwetzlmayr:
I do not consider this necessary. Feed detection should be sufficiently covered by the various ``<link rel="alternate"...`` elements in head. Also, might break existing sites by introducing yet another hard-coded URL segment.

Aeyoun:
The whole purpose of this pull request was to add _guessable URLs_ to Textpattern. You can tag `/feed` at the end of a URL on anything that looks like a blog, and it’s more likely than not that you’ll get a feed in return. I beleive that is valuable to support.

It also means you can more easily migrate to Textpattern from WordPress (and many other CMSes) without having to manually set up redirects for feeds.

da2x:
No interest?

philwareham:
@bloatware @Bloke Any thoughts on this?

Bloke:
Not a massive fan of this, sorry. We already have `/rss` and `/atom` endpoints hard-coded (which is bad enough) without introducing more.

If site admins feel strongly about having `/feed` also as some compatibility layer for their chosen syndication demographic, it's one `.htaccess` rule away to redirect to `/atom`.

But if this is a big issue for users, then fine. How important is it for regular folks having a guessable endpoint the same as WordPress? Are we talking a lot of people? Or edge cases? Aren't browser manufacturers and search engines trying desperately to phase out reliance on the URL anyway by hiding the address bar these days?

bloatware:
No idea. I'm absolutely indifferent to url prettiness and seo stuff. Not saying this is not important, just no idea.

philwareham:
Closing due to majority vote against. Thank you for the contribution though Daniel, appreciated.

PR #833, opened 2016-11-05
Add JSON for release and prerelease versions

petecooper:
JSON value for prerelease (beta, etc) should be `null` if there is no beta, and version number if if there is one.


philwareham:
Thanks again Pete, we will harnessing this during our migration away from RPC in future.

PR #840, opened 2016-11-10
Bloatware evaluate

bloatware:


philwareham:
Note to self: needs to be replicated in default theme repo too.

bloatware:
Sorry, I hope it wont move anymore.

PR #846, opened 2016-11-18
Ajax is caps case, not upper case

petecooper:


philwareham:
There's quite a few instances of that in the history doc. Might want to amend them all.

Bloke:
And there was me thinking it was an acronym for Asynchronous JavaScript And XML, hence I've always capitalised it. Learn something every day... thanks, Pete.

petecooper:
Google coined it here: http://adaptivepath.org/ideas/ajax-new-approach-web-applications/
Wikipedia call it Ajax: https://en.wikipedia.org/wiki/Ajax_(programming)

…I'm assuming of course the parser is part-powered by https://en.wikipedia.org/wiki/Ajax_(cleaning_product) but that's for another issue…

PR #849, opened 2016-11-19
Add Markdown versions of default article, plus typo

petecooper:
`summarise` -> `summarize` (see https://github.com/textpattern/pophelp/issues/24 for reasoning)

### DIFF
--- a/textpattern/setup/article.body.md
+++ b/textpattern/setup/article.body.md
@@ -1,0 +1,20 @@
+### What do you want to do next?
+
+-   Modify or delete this article? The [article list](siteurl/textpattern/index.php?event=list) is the place to start.
+-   Change this site's name, or modify the style of the URLs? It's all up to your [preferences](siteurl/textpattern/index.php?event=prefs).
+-   Get yourself acquainted with Textile, the humane markup generator which comes with Textpattern? The basics are [simple](http://textpattern.com/textile-sandbox). If you want to
+    learn more, you can dig into an [extensive manual](http://textpattern.com/textile-reference-manual) later.
+-   Be guided through your [Textpattern first steps](http://textpattern.com/textpattern-first-steps) by completing some basic tasks?
+-   Study the [Textpattern Semantic Model?](http://textpattern.com/textpattern-semantic-model)
+-   Add [another user](siteurl/textpattern/index.php?event=admin), or extend the capabilities with [third party plugins](siteurl/textpattern/index.php?event=plugin) you discovered from the central plugin directory at [Textpattern Resources](http://textpattern.org/?)
+-   Dive in and learn by trial and error? Then please note:
+    -   When you [write](siteurl/textpattern/index.php?event=article) an article you assign it to a section of your site.
+    -   Sections use a [page template](siteurl/textpattern/index.php?event=page) and a [style](siteurl/textpattern/index.php?event=css) as an output scaffold.
+    -   Page templates use HTML (HyperText Markup Language) and Textpattern tags (like this: `<txp:article />`) to build the markup.
+    -   Some Textpattern tags use [forms](siteurl/textpattern/index.php?event=form), which are building blocks for reusable snippets of code and markup you may build and use at your discretion.
+
+There is a host of [Frequently Asked Questions](http://textpattern.com/faq/) to help you get started.
+
+[Textpattern tags](http://docs.textpattern.io/tags/), their attributes and values are explained and sampled within the [Textpattern User Documentation](http://docs.textpattern.io/), where you will also find valuable tips and tutorials.
+
+There's also a crowd of friendly, helpful people over at the [Textpattern support forum](http://forum.textpattern.com/). Come and pay a visit!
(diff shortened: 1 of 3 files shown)


philwareham:
This has been superseded by work in https://github.com/textpattern/textpattern/pull/857, although the Markdown content here should eventually be pasted into the `setup/articles/` directory so please leave this PR open for now.

philwareham:
Now that the new setup structure has landed - we can revisit this. Shall we add markdown format to https://github.com/textpattern/textpattern/blob/dev/textpattern/setup/articles/articles.welcome.xml or park this for now?

philwareham:
Bump?

tecumsehmaverick:
Just a thought - and more likely than not, it is not a beneficial thought. But on the off-chance:

I really like the new xml schema that makss has created, and the new setup structure. 

I noticed in another thread about the issue with escaping json and the comment that the body and excerpt's raw source was not available.

And then the ongoing  textile vs markdown vs other text formatters. Plus the need sometime for plain text and even for those who might want straight (x)html(5).

What if the raw source was stored as xml/xhtml without text formatter specific markup, then transformed using xslt?

Along the lines of this? http://www.lowerelement.com/Geekery/XML/markdown.xsl

Granted, for the immediate issue at hand it would be simpler and more efficient to save a couple of variations (though I personally hate seeing the optional cruft add up).  But perhaps in broader scope of handling raw source and formatting, the need for text-formatter specific versions could be eliminated entirely?

Again, just a thought, and probably not overly helpful. :)

bloatware:
> What if the raw source was stored as xml/xhtml without text formatter specific markup, then transformed using xslt?

Where would this source come from?

tecumsehmaverick:
@etc.  I probably should not have made my suggestion since I'm mostly ignorant of how things are saved.

So risking sounding even more foolish

My thinking is running along this line:  my (mis?)understanding is that currently there are three copies of each body field (and 3 for each excerpt field) each time we save an article. One copy that is the raw source, once copy of formatted by textile, and one copy that is (x)html. Though as I write that I realize I don't know the difference between the raw source and the textiled source since we input it as textile.  I guess I assumed it was save as plain text.

It if is three copies, that seems like a potential inefficient use of database space. Even more if we end up with unlimited custom fields that also formatted text fields, each with three copies.  Thinking of unlimited fields +  the multilingual improvement, and if the 3 copy per is accurate, then we would have three copies for each field in each language.  That seems like it could add up in a hurry.

Thus my suggestion/query.  Would it be possible to rework how article are saved to work off of a single copy and transform as necessary.  

Specifically in answer to your question, would it be possible when an article is saved to translate those text boxes into a base plain text file with xml mark up as the raw source. Just one single copy tagged in xml.

ie. That any text formatter (Textile and Markdown being the specific examples) would be translated to a standard xml schema on save, and then available to be translated back.  

A potential advantage of this would be to free the content for pretty much any use and to be transformed using xslt in creative ways. 

Perhaps it might also allow text formatting choices to be made at the user level rather than at the textbox level.  If this users loves Textile, no problem. let them choose Textile. That one Markdown? Go for it.  Another wants some sort of pop up wysiwyg  - it should work as long as it was mapped to the xml schema.

But again, I realize I spoke when I don't really understand how everything works under the hood, and my ignorance is showing.

philwareham:
@tecumsehmaverick there's only two parts to a body or excerpt field:

1. The original text that the user enters in the textarea.
2. The HTML conversion of that text, using whichever text filter has been applied.

From what I gather, if a user selects 'Leave text untouched' and then saves the article. The original text and the HTML conversion are the same content.

tecumsehmaverick:
Thanks for the patient explanation Phil, and helping me with  my ignorance.

bloatware:
@tecumsehmaverick that was just a question, no worries :-)

> Specifically in answer to your question, would it be possible when an article is saved to translate those text fields into a base plain text file with xml mark up as the raw source. Just one single copy tagged in xml. ie. That any text formatter (Textile and Markdown being the specific examples) would be translated to a standard xml schema on save, and then available to be translated back.

That's (more or less) what Textile does: translate plain text to Textpattern-compatible HTML(ish). This HTML is then stored in `Body_html` field and served (after txp tags processing) to the client. The problems I can see with translating to XML are:

* Textpattern is not XML-compliant (tags in tags and so on), so storing articles with `<txp:tags />` inside can be problematic;
* When serving an article, one additionally has to translate XML to HTML, and this adds to runtime (server or client side).

> A potential advantage of this would be reduced space in the database, and freeing the content for pretty much any use and to be transformed using xslt in creative ways.

The translation Textile (say) to XML is not one to one: once an article is saved, you can obtain a different Textile source when you want to reedit it.

But I don't say it's not feasible, just don't know how.

tecumsehmaverick:
@bloatware  - Thanks.  That makes sense. I wondered about the added runtime and the type of hit it might cause. I didn't think about the tags within tags issue.  

As I said, my ignorance was being exposed. Thanks for the kind and patient reply.

bloatware:
I like the idea of "standardized" `Body`, quite natural. Just am too ignorant to implement it!

petecooper:
Looking through the comments and the work undertaken since this PR was raised, and taking into account the branch conflict, I'm going to close this. It's straightforward to port the existing article body and article into another format, so as and when that's necessary it can be done. I'm going to close this since it's making the place look untidy. Thank you for your input, and my apologies for radio silence in recent months.

philwareham:
Cheers Pete. And good to hear from you!

PR #855, opened 2016-12-06
Simplification atom/rss feed

makss:
Small optimization and code simplification.
* Minus one sql query (prefs `lastmod` already loaded)
* Minor code sync between atom.php and rss.php

### DIFF
--- a/textpattern/publish/atom.php
+++ b/textpattern/publish/atom.php
@@ -85,8 +85,6 @@ function atom()
     ob_clean();
     extract($prefs);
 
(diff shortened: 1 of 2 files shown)


bloatware:
Good call, but I'm not sure `strtotime($lastmod)` and `UNIX_TIMESTAMP(val)` will always be equal, unless `$lastmod` is somehow re-time-zoned before extraction. If `$lastmod == val` then we should use `strtotime($lastmod+' DB Server timezone')`, or am I paranoiac?

bloatware:
Actually, this request also fixes `lastmod` timestamp discrepancy when PHP and DB servers are in different timezones. Merging, thanks @makss.

Issue #575, opened 2015-10-31
Drop the prefs_id column in txp_prefs


vanmelick:
That column isn't used for anything. All prefs have a prefs_id with value 1.
Depends on #573

Bloke:
Axing it gets my vote. Never seen the point of it.

Bloke:
@vanmelick Do you have a patch for this lurking anywhere that you could apply?

philwareham:
This has been done, yes? I don't see that column in my 4.7 installs. We can close then?

makss:
Yes, can close it.

philwareham:
Great, closing, thanks.

Issue #777, opened 2016-09-26
Add check(or AutoFix) TxP core preferences


makss:
I propose to add a check basic core preferences (`txp_prefs`) to `Diagnostic` tab.
Can make and AutoFix missing preferences.

All you need we already have:
1. Move array `$prefs = array_merge_recursive(array(...));` from `/textpattern/setup/txpsql.php` to new file `/textpattern/lib/prefs.php`
2. Include `/textpattern/lib/prefs.php` in `/textpattern/setup/txpsql.php`
3. Include `/textpattern/lib/prefs.php` in `/textpattern/include/txp_diag.php`
4. Create new function AutoFix in `txp_diag.php`:  Create prefs if not exist

makss:
Can do the checking and recovery after login to the admin interface with admin privileges.
This will reduce the amount of code in `/textpattern/update/` folder and in setup stage too.

Bonus: We work with the core prefs will be in one place vs many files(update/setup)/functions(autocreate in function)

bloatware:
Let's postpone the discussion to 4.7.

bloatware:
@makss would you mind submitting a patch?

Bloke:
@makss And if you do submit a patch, would you please think about the following:

As part of the Themes implementation, some themes might require certain values for existing prefs or to install new prefs. So, during theme import/export if there was a core API method to set prefs from a file bundled with the theme, that'd be brilliant. This seems to have some crossover with this issue.

Bundling of prefs with themes was not slated to hit 4.7.0, but if we have a mechanism to import them available in core, it might make it viable to roll in support alongside the basic themes implementation. Thanks!

makss:
Another idea was: it would be good to make the structure of sql tables in a separate file, for example `/textpattern/lib/tables.php`

```
$txp_tables = array();

$txp_tables['txp_category'] = "
    id          INT          NOT NULL AUTO_INCREMENT,
    name        VARCHAR(64)  NOT NULL DEFAULT '',
    type        VARCHAR(64)  NOT NULL DEFAULT '',
    parent      VARCHAR(64)  NOT NULL DEFAULT '',
    lft         INT          NOT NULL DEFAULT '0',
    rgt         INT          NOT NULL DEFAULT '0',
    title       VARCHAR(255) NOT NULL DEFAULT '',
    description VARCHAR(255) NOT NULL DEFAULT '',
    PRIMARY KEY (id)
";

$txp_tables['txp_css'] = "
[...]

PR #857, opened 2016-12-06
AutoFix TxP core prefs

makss:
Issue #777 , #575 

Prefs:
---
* After user logon complete check and AutoFix Core/Theme, Global/Private prefs. [Crash Test](https://github.com/textpattern/textpattern/pull/857#issuecomment-269245590)

* TxP Core Global and Private prefs defined in file
`/textpattern/vendors/Textpattern/DB/Data/core.prefs` *(AutoFix on)*

Tables:
---
* The structure of the sql tables moved into separate files.
`/textpattern/vendors/Textpattern/DB/Tables/*.table`

* Initial mandatory data moved in separate files.
`/textpattern/vendors/Textpattern/DB/Data/*.xml`

Themes:
---

* Theme prefs:
  * `/data/core.prefs`    - Allow override some core prefs. Used only in setup theme. *(AutoFix off)*
  * `/data/theme.prefs`   - Theme global and private prefs. *(AutoFix on)*
     * global  - Used in setup and for AutoFix prefs.
     * private - Will be created after user login
.

* Additional data.
`/textpattern/setup/data/*.xml`

* Articles here
`/textpattern/setup/articles/*.xml`

```
<?xml version="1.0" encoding="utf-8"?>
<resources>
<articles>
<article>
    <title>Welcome to your site</title>
    <section>articles</section>
    <category>hope-for-the-future</category>
    <category>meaningful-labor</category>
    <body format="textile"><![CDATA[
h3. What do you want to do next?
...
]]></body>
    <excerpt format="textile"><![CDATA[
This is a Textpattern article excerpt.
]]></excerpt>
    <comment>
        <name>Donald Swain</name>
        <email>donald-swain@example.com</email>
        <web>http://forum.textpattern.com/</web>
        <message><![CDATA[<p>I enjoy your site very much.</p>]]></message>
    </comment>
    <comment>
        <name>Angry Birds</name>
        <web>http://forum.textpattern.com/viewforum.php?id=5</web>
        <message><![CDATA[<p>Welcome article - Second comment</p>]]></message>
    </comment>
</article>

<article>
    <title>Second article</title>
    <section>articles</section>
    <body><![CDATA[
<h3>HTML body content</h3>
]]></body>
    <excerpt><![CDATA[
<h3>HTML excerpt content</h3>
]]></excerpt>
</article>
</articles>
</resources>
```
Available optional fields for the `article`:

* `status` - default - `Live`
* `keywords`
* `description`
* `annotate`
* `annotateinvite`
* `custom_1` - `custom_10`

`comment`:

* `email` - if not set then compose email as `stripSpace(comment.name).'@example.com'`


textpattern/lib/txplib_misc.php
```diff
@@ -6747,3 +6747,25 @@ function check_file_integrity($flags = INTEGRITY_STATUS)
...
+
+function check_prefs_integrity()
+{
+    global $prefs;
+    $prefs = get_prefs();
+    include txpath.'/lib/prefs.php';
+
+    foreach ($default_prefs as $event => $event_prefs) {
+        foreach ($event_prefs as $p) {
+            if (! isset($prefs[$p[3]])) {
```
Bloke: Can we not use `set_pref()` or `create_pref()` here instead of a raw insert? We could then take out any (all? most?) `doSlash()` calls in `prefs.php` because we get it done for free in those functions.

Since we're not setting per-user prefs here (so `$txp_user` doesn't matter), and `txplib_db.php` is included in your patch, we might be able to get away with using the prefs API functions. Or have  missed something?

makss:
@Bloke, Thanks, you are right, to use the prefs API will be more convenient.

makss:
The second step, welcome to the test.  ;-)

textpattern/setup/txpsql.php
```diff
@@ -51,130 +51,130 @@
...
 
-require_once txpath.'/lib/IXRClass.php';
-$client = new IXR_Client('http://rpc.textpattern.com');
 
-if (!$client->query('tups.getLanguage', $blog_uid, LANG)) {
-    // If cannot install from lang file, setup the English lang.
-    if (!install_language_from_file(LANG)) {
-        $lang = 'en-gb';
-        include_once txpath.'/setup/en-gb.php';
+function setup_txp_lang($lang)
```
Bloke: Don't worry too much about refactoring any of this RPC stuff. The RPC reliance is (hopefully) disappearing in this release and we'll install lang strings from files bundled in core. See the languages-inclusion branch for more.

makss:
* For articles enough fields `title`, `category`, `body`, `excerpt` or add some other field? `Section` and `id` is taken from the filename.

* Is possible to add user preferences autocreation after user login. Describe it as `$default_prefs_user` and drop function `new_user_prefs()`.

* I do not know how better to describe the preferences for the theme, I will be happy to advice.

Bloke:
> For articles enough fields title, category, body, excerpt or add some other field? Section and id is taken from the filename.

I'm not wildly enthusiastic about the articles implementation, tbh. The `.textile` suffix doesn't entirely match its content, and the chosen delimiters for the fields (the three equals symbols) kind of clash with Markdown and are therefore confusing.

Also,what's the benefit of specifying the ID in the filename? If we are are thinking of offering 'repair tables' or some other feature based on the structure, we won't be re-inserting the data, just in case the customer has changed article ID 1. Therefore the only time this article is useful is during setup, and we know it'll be ID 1 because it must be done on an empty table. So why do we need the ID?

Is it just in case we (or someone who wants to pre-install an environment with a bunch of articles in future) can insert multiple articles in a specified order? That's fine if so. As long as it will work reliably if there are, say, 12 articles. In that case the order of insertion will be: 1, 10, 11, 12, 2, 3, 4, 5, 6, 7, 8, 9. Presumably MySQL figures that out and it doesn't break the autoinsert numbering?

> Is possible to add user preferences autocreation after user login. Describe it as $default_prefs_user and drop function new_user_prefs().

That would be much better! Please do that if you can, it centralises all the prefs creation code.

> I do not know how better to describe the preferences for the theme

I was thinking of a standard definition file format, like `my-theme.prefs` which could be read using a similar system to your `get_files_content()` function, and the list of prefs inserted. Some way to distinguish which are global prefs and which are per-user would be handy too. That way, the function to install global prefs can be called when the theme is installed, and the per-user theme prefs will get added when a user authenticates (i.e. visits a new admin panel) and has enough privs to check the prefs integrity. We could use JSON as a format?

```
{
   "my-pref": {
      "value": "foo",
      "event": "blue_theme",
      "html": "text_input",
      "type": "hidden",
      "position": "10",
      "private": "true",
   },
   "my-other-pref": {
      "value": "bar",
      "event": "blue_theme",
      "html": "text_input",
      "type": "hidden",
      "position": "20",
      "private": "false",
   }
}
```

You could maybe extend that idea to the core prefs, but I don't know how you would inject PHP variables safely (e.g. `$setup_comment_invite`).

makss:
> I'm not wildly enthusiastic about the articles implementation, tbh. The .textile suffix doesn't entirely match its content, and the chosen delimiters for the fields (the three equals symbols) kind of clash with Markdown and are therefore confusing.

I also do not particularly like it, but we need also to start somewhere. ;-)
I would like to pick up such separators, which would not disturb edit textile(Markdown).

> Probably a better system out there somewhere. Maybe the way Jekyll does its metadata headers?

We have two text fields `body`, `excerpt`. I can not quite imagine how to do that with Jekyll

> Also,what's the benefit of specifying the ID in the filename?

Honestly - it's my laziness.  ;-)   Probably have a container(article) include comments that are linked to it.
Probably the file name will be something like:  `section.some_not_used_text.article` - or another extension?

> I was thinking of a standard definition file format, like `my-theme.prefs`
We could use JSON as a format?

Good idea, is not very convenient to edit, but it is a common standard.
It would be nice to add a table `txp_prefs` information field `owner` (or `source`) to be seen therein whose this preference: `core`, `my-theme`, `abc_plugin`

Is still possible to use a json format for presenting data in sql tables.
`/textpattern/update/structure/*.data`
`/textpattern/setup/data/*.data`

> You could maybe extend that idea to the core prefs, but I don't know how you would inject PHP variables safely (e.g. $setup_comment_invite).

More will need to somehow describe some sql functions or values: `now()`, `NULL`

philwareham:
@makss how about an XML file for this task? We are thinking of using that format for (pop)help files so there will be precedent elsewhere in the system. Just a thought.

makss:
@philwareham  It will be convenient to edit xml file in which the textile?
I agree that the separator `===` not the best, but it does not prevent the textile editing. You can edit a [file](https://raw.githubusercontent.com/makss/textpattern/f5e7c36f1d463d8287e12655b53cf003e5e3dfea/textpattern/setup/articles/1.articles.textile) (as article in textpattern) and immediately see the result.

philwareham:
Something like this? 
[example.xml](https://github.com/textpattern/textpattern/files/642387/example.txt)

makss:
> Something like this?  example.xml
```
<code>&lt;txp:excerpt /&gt;</code> tag.
```
;-(

philwareham:
> ;-(

Yes, but ideally we should move away from Textile to HTML for this, since in theory if Markdown core support materialises then in your example we would have to also provide a Markdown version of that Textile file. But that throws up how we convert that HTML back into Textile/Markdown I guess. Hmmm.

Bloke:
If we're going XML, this is better:

```
<?xml version="1.0" encoding="utf-8"?>
<resources type="article">
  <title>Welcome to your site</title>
  <category>hope-for-the-future</category>
  <category>meaningful-labor</category>
  <body format="textile">h3. What do you want to do next?
...
  </body>
  <body format="markdown">### What do you want to do next?
...
  </body>
  <excerpt format="textile">This is a Textpattern article excerpt.
Article excerpts are optional but can be used effectively to
summarise an article's content, for example with Textpattern's
built-in article syndication and with the @<txp:excerpt />@ tag.
  </excerpt>
</resources>
```

Notice we use specific tags and attribute modifiers instead of generic `<string>` tags. This adds flexibility, reduces duplication and typing, and also permits niceties like specifying the text in multiple markup languages.

Note also that a resource has a `type="article"`. Makes it easy to port the concept to other content types in future. Just sayin'.

Bloke:
> It would be nice to add a table txp_prefs information field `owner`

Yes, please! We have that column for the txp_lang table so it makes sense to introduce one for prefs too. I was thinking of reusing `event` for that, but it's better to keep that for sub-divisions and introduce a new top-level `owner` which can collect related prefs together.

Also, see #575 which is kind of related to all this prefs work. So if you happen to be in the mood and wanted to tackle that Issue as well, feel free :-)

philwareham:
@bloke OK that XML file looks good to me. Wondering if all this stuff can transfer nicely to the themes work at some point? Would be nice to have different startup content depending on the theme you install (like some dummy content on a news-type site or corporate-type site).

Bloke:
@philwareham I hope so!

philwareham:
FYI - I've fixed the merge conflict

makss:
@philwareham  Thanks

makss:
Add to first post the file format description of `articles/*.xml`.
Anything else to add/remove from xml?

Bloke:
Something's rumbling at the back of my brain here. Please bear with me while I dump the contents of my head here.

Since the scope of this patch is rapidly expanding, what we seem to have stumbled upon is a generic way to import content. Thus, anyone could, in theory, export article and potentially other content metadata from any system into our burgeoning XML / JSON format and we could offer a way to load it into the DB. That's pretty powerful and essentially offers custom functionality we lost when we axed the (broken) 'import' feature in 4.6.0. "Other systems" could be another Txp instance, another CMS, etc.

With that in mind, where does this facility fit? Obviously it's useful for installation and (perhaps) upgrades, as demonstrated by this patch. Presumably also, for loading content alongside Themes when that gets merged in -- although loading assets without trampling on existing content would be a challenge.

So, does this part of the patch deserve to be made into a proper class in our `vendors\Textpattern` namespace? That way, we can use it outside of the setup, and all the new functions are encapsulated. The setup utility already has access to the class loader so we can use it. And, crucially, plugins would then be able to access it more readily so someone could write an import/export plugin if they wished.

This facility would be even more powerful if it raised callbacks so that plugins could do more too, e.g. before insertion, raise a callback_ref and pass the proposed data out so the payload can be manipulated before it's inserted.

Just shooting ideas out at the moment and I don't wish to burden makss any further with this spiralling patch, but does it make sense to bundle some parts of this patch up into a class instead of introducing more global namespace functions?

textpattern/setup/txpsql.php
```diff
@@ -125,7 +127,17 @@
     $xml = simplexml_load_string($data, "SimpleXMLElement", LIBXML_NOCDATA);
     foreach ($xml->article as $a) {
         $article = array();
-        $article['Section']   = empty($a->section) ? $section : $a->section;
+        $article['section']   = $section;
+        $article['status'] = 4;
```
Bloke: Can you use `STATUS_LIVE` here, please? And anywhere else where status codes have used magic numbers. Thanks.

makss:
I did not want to dive so deep. ;-)
Just try to get to the preferences and a little to clean up the installation process. The patch is strongly inflated, we have to somehow limit it.

Ability to import / export - large, it is advisable to leave them for plugins. Plugins can be updated more quickly, to enable them to support a larger number of other cms. They can also be run on older versions of Textpattern.

Can add the base class in the `vendors\Textpattern`, but unfortunately I do not know very well object oriented programming and I can not do it.

Bloke:
I appreciate that the patch is suffering from feature creep. Didn't expect it to encompass so much at the outset! But now that it has, we could either:

a) Proceed as I mentioned above.
b) Split the functionality up into different PRs (if possible, though it seems very entwined with the import/export stuff now).
c) Carry on as we are and worry about OO later.

The major downside to (c) is that we introduce more global-scope functions and then immediately deprecate / remove them when the code moves to a class a few days/weeks later (or when Themes get merged). I suppose at this early stage in the 4.7.0 lifecycle we could just remove the functions because nobody outside core is likely to have used them yet. But the longer they stay in the codebase, the more likelihood there is of someone using them and then it becomes more difficult to remove them.

I completely agree about plugins being the vehicle to offer import/export. It just struck me that the systems and functions makss has put in place would be 75% of the functionality required to implement such a thing, so to add a few choice callback hooks and bundle up that part into an "official" class now would make perfect sense. The Themes feature could then use it if necessary to offer content I/O (in a future release) alongside template I/O (planned this release), and in the meantime plugins could spring up to cover the gap.

As the concept evolves, there's nothing to stop us expanding the XML/JSON scope to handle unlimited categories/tags, and rolling in support for unlimited custom fields, as and when those things enter core.

I could look at the OO side if you like? Though I'm not sure if it's possible for me to commit to the patch, nor how to go about it if so. Do I have to fork a PR to do it into my own branch? Or we could just go with option (c) and we can rejig things when it's merged to core.

philwareham:
I think option C is fine as long as the code doesn't hang around too long before moving to OO. As long as you let plugin authors know in the forum I think we are safe.

makss:
It seems that's all.
Can test, merge, modify, OO, etc.

makss:
Complete check and fix prefs.
(fields: `type`, `event`, `html`, `position`).

Crash tests:
---

1. Drop all prefs
`truncate table txp_prefs`
After logon Core and Theme, Global and Private prefs will be restored.

2. Random values in fields: type, event, html, position.
`update txp_prefs set type=rand()*99, position=rand()*99, event=rand(), html=rand()`
After logon Core and Theme, Global and Private prefs will be restored.

.

If all is well, can remove from folder `/textpattern/updates/` all references to `safe_update('txp_prefs'...`

makss:
`Drop the prefs_id column in txp_prefs` commit close Issue #575

makss:
I will try to structure changes in this patch.

It took first extract the data from the php code, so the patch so the gap. Now I will try to integrate this patch into the mainline. I think that further expansion should be done by individual patches. (import/export/themes/etc)

1. Extract data from php file `/textpattern/setup/txpsql.php` to folders:
`/textpattern/textpattern/update/structure/`
`/textpattern/textpattern/setup/articles/`
`/textpattern/textpattern/setup/data/`


2. Completely rewritten DB installation process. File `/textpattern/setup/txpsql.php`

Have been made data loaders:

File `/textpattern/setup/txpsql.php`
* function article_import() - This function is only used for setup. It fully carries out its work. Its further expansion left the following patches.

File `/textpattern/lib/txplib_misc.php`
* function get_prefs_default()
* function get_prefs_theme() - In the future, need to move in `/vendors/Textpattern/Admin/Theme.php` or some else.
* function get_files_content()

File `/textpattern/lib/txplib_db.php`
* function make_sql_set()


3. AutoFix TxP Core/Theme (Global/Private) prefs 

File `/textpattern/lib/txplib_misc.php`
* function check_prefs_integrity() - function is called after the user login.

4. Removing unnecessary function `new_user_prefs()` - This is done function `check_prefs_integrity()`.

5. Removing unnecessary code from `/update/` folder. This `txp_prefs` work has already performs the function `check_prefs_integrity()`

6. Drop the prefs_id column in txp_prefs


No other changes were made.

philwareham:
Are you all happy to submit this for testing and merging?

Bloke:
I was going to test it last night but forgot, sorry. Will try tonight and go from there.

Bloke:
@makss Sorry for the delay. This is really great stuff. I think we'll be able to migrate this to a more generally available class-based system under our namespace at some point in future. We could then use that to open up the ability for plugins to do content importing from files. And maybe even for other types of asset, too.

I might even do it for the next phase of the Themes work.

A few things to note:

* Can you please merge the current `dev` branch into the patch please before they drift out of sync? I did it locally on my copy today and it fast-forwarded nicely.
* The `default_publish_status` pref is a per-user pref. It's defined as such in `/textpattern/update/structure/core.prefs` but on a clean install, it's created as a global pref. This is probably because -- at setup time -- there is no `$txpuser` set, since you're not logged in. That's why the `new_user_prefs()` function used to exist, so the per-user prefs could be created at account creation time, because the user name is known at that moment. Aren't per-user prefs deferred to being created at login time if they don't exist? Isn't that one of the goals of the patch?
* Could we benefit by converting the JSON documents (e.g. `/textpattern/setup/data/txp_category.json`) to a similar XML format used for articles? It might be slightly more wordy, but would there be less escaping required? I'm thinking longer-term here: if we allow article content to be imported in XML format, why not links and categories and sections and ... ?
* Is the inclusion of the `en-gb.php` file in the `setup` directory a temporary measure? We already have all the strings available in the `lang` directory (albeit in `.txt` format) and you can extract the lastmod time from the first row of that file. So why duplicate them all in the PHP file? The advantages to the .txt format: a) no escaping required for apostrophes. b) in Txp 4.7.0 _all_ language files will be bundled in the `lang` directory, and I don't think it's a good idea to duplicate each one into a PHP file as well. Can the patch read the `lang/*.txt` files instead?

This is a seriously awesome upgrade to the way we handle installation and updating, and it's going to open up a lot of possibilities when we extend it to be available to plugins. Thank you for all the hard work!

makss:
> The default_publish_status pref is a per-user pref. It's defined as such in /textpattern/update/structure/core.prefs but on a clean install, it's created as a global pref.

ops...  It's my bug, fixed.

> Can the patch read the lang/*.txt files instead?

Done. (used `Textpack\Parser.php`)
.

json vs xml - I think about it a little later.

makss:
@Bloke  Can I ask you to put some files from the branch `languages-inclusion` to `dev` branch?
Files:
```
textpattern/lib/constants.php
textpattern/vendors/Textpattern/L10n/Lang.php
textpattern/vendors/Textpattern/Textpack/Parser.php
```

I would like to change function `Txp::get('\Textpattern\L10n\Lang')->install_file($lang)` (make use `Textpack\Parser.php`) and use `L10n/Lang.php` in my patch for `load lang/*.txt` files in setup, otherwise we will have a code conflict. ;-)

Bloke:
@makss I could pull those files over to `dev` and them manually merge `dev` back into `languages-inclusion` and fix the (non-existent) conflicts if it'll help you out.

Would you then also link `install_languages_from_file()` in your patch to use the new class and add the deprecation note, or shall we leave that bit for when `languages-inclusion` lands in `dev`?

makss:
Trying to apply object-oriented programming. Introduce `DB/Core` and `Import/TxpXML` class.
Tables structure, data, core prefs, some functions moved to `/vendors` tree.

philwareham:
@makss let us know when the PR is ready for testing and possible merge. Cheers!

Bloke:
This is super awesome, makss. I'm happy to merge it but before I do, what are you intending to do with the additional content such as `article_mixtype_test` and the additional assets it creates (such as the test categories, links, css, forms, pages, and `articles2` section?) Shall we merge it all and then remove them afterwards, or remove them prior to merge?

We could do with that article being kept somewhere as an example to show off the powerful XML format you've created. I'm just not sure where to put it in a default install environment. Maybe we just document it somewhere for now? What do you think?

I think this'll be amazing for offering migration functionality from other CMSs or databases or spreadsheets because people can just output it in the required format and inject it straight into Txp. And of course for Themes file system support. Fab!

makss:
Yes, of course, I removed test cases and data so that they do not interfere with the main branch.
Documentation can be written later.  In order not to lose, the example file `theme.prefs`
```
{
    "hive_test1": {
        "val": "It's theme Global pref",
        "type": "0",
        "event": "theme",
        "html": "text_input",
        "position": "0"
    },
    "hive_test2": {
        "val": "It's theme Private pref",
        "type": "0",
        "event": "theme",
        "html": "text_input",
        "position": "0",
        "private": "1"
    }
}
```
Migration and import/export should probably be done already in the following patches.

PR #884, opened 2017-02-11
file panel – date sort in file list

jools-r:
Two mini patches
– clicking on date column head in the file panel currently switches to the image panel!?
– add support for sorting the file list by date (using created date)


philwareham:
Tested and merged. Thanks for the contribution!

PR #889, opened 2017-03-19
Fix feed warning + formatting output

makss:
Fix warning

`http://domain.com/?rss=1&area=link`

```
Tag error:  -> Notice: Undefined variable: uLastMod while parsing form <strong>None</strong> on page <strong>default</strong>
```

and formatting feeds output


bloatware:
Thanks @makss

PR #893, opened 2017-04-03
Global attributes

bloatware:


bloatware:
Time to test

phiw13:
Yay…

Hmm. You broke `zem-contact-reborn` :-(

The plugin has a `label` attribute for every tag. It now is outputted twice (2x). Sample below is almost verbatim copy/paste from the plugins help file. No messages in debug mode, btw.

Input in my TXP form:

```
<div><txp:zem_contact_email break="" class="" label="Email" name="email" placeholder="Your email address" /></div>
```

Actual output (note the string `Email` followed by a `<br>` at the very beginning):

```
<div>Email<br>
<label for="email" class="required-field">Email</label><input class="required-field" id="email" name="email" type="email" maxlength="100" form="zcr6ec7bda7a5484136123864ec3f57f736" placeholder="Your email address" required></div>
```

Expected output

```
<div><label for="email" class="required-field">Email</label><input class="required-field" id="email" name="email" type="email" maxlength="100" form="zcr668bc7134bad24f69f8c56ea42482d23" placeholder="Your email address" required></div>
```

So far it is the only thing that ”broke” on  a (local) mirror copy of a real live site with all its historical bagage. I haven’t done any edits (yet) to try to use any new functionality.

Bloke:
@phiw13 ZCR uses its own version of `lAtts()` and implements its own version of "global" attributes so there's every chance the plugin is at fault here. Once I've had a chance to grok Oleg's work more thoroughly I'll probably make a 4.7.0.0 branch of the plugin and work there.

@bloatware Is there a better way the plugin can register its own global attaributes now? It has a set of attributes for HTML 5 that work across all the tags, so I hacked in support for this at the `lAtts()` level. There may well be a better way for me to do that now that global attributes are here, so any advice welcome. Thanks.

phiw13:
@Bloke – yeah I know that ZCR uses its own stuff. And I didn’t really accuse Oleg of breaking anything, sorry if it came over like that. It was more a thong-in-cheek kind of answer, as ZCR was the only thing that broke across that test site.

bloatware:
Well, every commit is reversible :-) I'm still not fully convinced of the global attributes utility, need others advice. Plugins can be patched, sure, but is it worth it?

> Is there a better way the plugin can register its own global attributes now?

I've tried to implement it without the need for plugins to register their global attributes (think of old unmaintained plugins). Anyway, if a plugin registered, say, `label` as global, it would apply to all tags that do not support their own `label` attribute.

The most straightforward way to patch ZCR (and other plugins that bypass `lAtts()`) is to call
```
lAtts(array('label' => ''), $atts, false);
```

somewhere in `zem_contact_lAtts()` or other universally processed ZCR function. This will unset the global `label` attribute.

PR #901, opened 2017-05-02
Avoid undefined variable notices

rimas-kudelis:
This fixes an issue reported in https://forum.textpattern.io/viewtopic.php?id=47793

### DIFF
--- a/textpattern/lib/txplib_publish.php
+++ b/textpattern/lib/txplib_publish.php
@@ -188,6 +188,8 @@ function getNeighbour($threshold, $s, $type, $atts = array(), $threshold_type = 
         return $cache[$key];
     }
 
+    $expired = $id = $time = $keywords = $sortdir = '';
+
     extract($atts);
     $expired = ($expired && ($prefs['publish_expired_articles']));
     $status = isset($status) && intval($status) == STATUS_STICKY ? STATUS_STICKY : STATUS_LIVE;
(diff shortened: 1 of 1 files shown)


philwareham:
@bloke @bloatware can you check this over and merge if suitable? Cheers all!

bloatware:
Not sure. The issue is caused by placing `<txp:link_to_next />` before `<txp:article />` (who sets the necessary variables). This patch would remove the warnings but not the possibility of a wrong `<txp:link_to_next />` in case `<txp:article />` uses (say) a non-default sort order.

rimas-kudelis:
I don't think that using potentially undefined variables is a good thing, and expectations that users will only use tags in a certain sequence are obviously prone to failure.

That said, I'm not claiming my fix is the best approach to tackle the problem, since I'm not acquainted with txp well enough (if at all). The ball is yours now, so proceed as you see fit. ;)

Bloke:
> expectations that users will only use tags in a certain sequence are obviously prone to failure.

Yes, it's a long-standing issue in Textpattern, with no easy (read: backwards-compatible / performance-maintaining / robust) solution. It can be fixed, but none of the current solutions are optimal.

Issuing PHP warnings when tags are used in the wrong order is not acceptable though, so the warnings should be removed, even if the solution doesn't (yet) address the underlying issue of tag order.

bloatware:
I certainly agree the code robustness is essential and thank you for the submission, it will serve (after refactoring default values). But I wouldn't like that doing it obfuscates the underlying issue, that's all.

bloatware:
Thanks @rimas-kudelis!

Issue #408, opened 2014-10-04
X-Frame-Options is deprecated


ghost:
_From [jukka.m.svahn](https://code.google.com/u/jukka.m.svahn/) on December 09, 2013 13:17:12_

X-Frame-Options is being deprecated in Firefox in favor of Content Security Policy. http://www.w3.org/TR/CSP Which is standard candidate. We should migrate to it and leave the old X-Frame-Options as fallback for older browsers. E.g.

header('Content-Security-Policy', "default-src \* 'self' 'unsafe-inline' 'unsafe-eval'; frame-src 'self'");

_Original issue: http://code.google.com/p/textpattern/issues/detail?id=404_

vanmelick:
I know it's an example, but...
* frame-src 'self' is deprecated; should be replaced with child-src 'self'... but as I understand it, that means the page can only use iframe/frame children that come from the same server, which would break youtube movies, which often (typical?) load via an iframe.
* setting `default-src` as in the example doesn't seem to do much, because of the asterisk in there and allowing unsafe-inline / unsafe-eval. If the default is set like that, wouldn't it be better not to set a default and only send the header if the user defines it in config.php... and in that case, wouldn't it be just as easy for the user to call the header function from config.php?
* X-Frame-Options: SAMEORIGIN my be deprecated, but its proposed replacement - the directive `frame-ancestors` in CSP level 2 is marked as experimental / not for production use at https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives#frame-ancestors

Bloke:
No idea what the correct solution is here. @vanmelick: care to propose one?

philwareham:
Bumping - @vanmelick @rwetzlmayr @bloatware do you know of a solution for this?

bloatware:
Sorry, I don't understand the issue itself, the [original issue](http://code.google.com/p/textpattern/issues/detail?id=404) link points to something seemingly unrelated.

petecooper:
Browser support for CSP: http://caniuse.com/#feat=contentsecuritypolicy
CSP policy directives: https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives
And some OWASP for bedtime reading (search for 'deprecated'): https://www.owasp.org/index.php/Clickjacking_Defense_Cheat_Sheet

bloatware:
We already ship it with `.htaccess`, but this works only (?) for apache servers.

[...]

PR #916, opened 2017-07-06
Set Content-Security-Policy header from php

makss:
Issue #408
* Reduces the mix of headers (.htaccess/php).
* It is possible to set the header under the nginx and other web servers.


philwareham:
Great, thanks! Another issue closed.

PR #918, opened 2017-07-10
Stable test of clean URLs

makss:
Clear the output buffer before running the test.

https://forum.textpattern.io/viewtopic.php?id=47898


Bloke:
Good idea, thanks makss.

PR #919, opened 2017-07-16
Set CSP headers from one location

makss:
* Cleaned the code repetition.
* Use pagetop() to set the CSP headers in `async` mode


philwareham:
Thanks dude!

Issue #887, opened 2017-03-17
Consider Moving two Article preferences to either above or below Body and Excerpt fields


bici:
See this discussion https://forum.textpattern.io/viewtopic.php?pid=304837#p304837
I would consider a toggle to On or Off On to see te options and then off. Could be set by user to always on by default to off

philwareham:
Like [this?](http://design-patterns.textpattern.io/mockups/content-article.html)

bici:
i think that looks marvellous! and not overly intrusive. thanks

philwareham:
@Bloke over to you then!? ;)

tecumsehmaverick:
Just another perspective, fwiw.

I have been aware of those options were, so I do not have the same felt need for them to have great visibility.  I personally think it is more elegant, and a cleaner UI to have options on the side, and just content fields in the middle (as much as possible).

That said, it's not a huge deal if they are moved.  bici's mock up isn't bad looking either. Just not my first choice.

What for me would be a bigger issue is the upcoming unlimited custom fields.   Each field will have the potential for a markup language as I understand it?  So if I add several more fields, if we are consistent we now have extra options with each field?  That would be clunky imo.

As I said, just another perspective.

philwareham:
@tecumsehmaverick those of of course valid points also. Regarding custom fields there would need to be some tie-up between custom fields (textareas) and their markup preference though, so they'd need to be paired together like the above example. I agree it's extra cruft that cannot be hidden inside a side panel though, and I like less UI not more.

The mockup is just for discussion hence why I've differed to @Bloke.

bici:
Thanks tecumsehmaverick - having used EE for years with unlimited custom fields each with formatting choice,  it is not a burden at all in terms of screen usage. In the other CMS one can toggle other custom fields to be open/visible by default or to only open when needed;  easily manageable. But i agree with you a UI is very personal. Too minimal in my view with an important Textual option, way off off to the side with no visible clue as to what it  belongs/affects is not something that i ever used...it was unclear to me what that option was meant for!   Now that i have used it is perfectly clear. but overall in my opinion it belongs CLOSE to the TEXT fields where a user is writing.

Bloke:
@tecumsehmaverick That's a good point. The plan is indeed to have markup per field -- at least for the text-based ones, and it will get... cluttered if we follow the same layout model there.

[...]

Issue #292, opened 2014-10-04
Improve (pop)help support


ghost:
_From [stefdawson](https://code.google.com/u/stefdawson/) on October 15, 2012 13:59:08_

The current help topics from the RPC server return complete markup and styling which makes it difficult to embed in Textpattern. It's also a barrier for plugin authors to write their own help topics using the new admin_help callback.

Propose adding a JSON-esque return path for RPC help topic requests, triggered by adding &type=json to the help URL. Returned data might include:
- Title
- Language code (which may be the requested lang or the default lang if help topic not yet translated)
- Raw (currently Textile) description which may be locally passed through a Textfilter (not sure if this is useful)
- Cooked HTML description suitable for direct injection into a markup template

As well as reducing traffic, help topics can then be built and styled locally, permitting modal and other implementations. Plugin authors could then just replace core URL paths with their own endpoint to serve similar JSON objects and have custom help topics rendered as per the current in-force theme.

See http://forum.textpattern.com/viewtopic.php?pid=259991#p259991

_Original issue: http://code.google.com/p/textpattern/issues/detail?id=288_

philwareham:
Current status is that @petecooper is batch converting the English (GB) pophelps at https://github.com/textpattern/pophelp into raw HTML. We will then clean-up the output and integrate this into work on https://github.com/textpattern/textpattern/issues/838.

petecooper:
See also: https://github.com/textpattern/pophelp/issues/32 for completeness.

makss:
Please check the work of popHelp, this functionality is already available.

Bloke:
This is fantastic, makss. Thank you. The popup is so much nicer than the separate window, and it works beautifully in other languages too (I gave it a quick test with a `fr_pophelp.xml` file).

The pref is a great idea too. One thought: in future, are we thinking of offering pophelp in a variety of formats? If so, would it make sense to rename the pref now and (maybe) use a select list here instead of a yes/no radio? Then we could more easily extend the system in 4.8 and beyond. For example:

Help system
* None
* Pophelp
* Slidey drawer thingy
* Menu
* ...

As it stands, if we define those as constants in future, we'll need to ensure that 'pophelp' = '1' so that people who have set the current 'use pophelp' yes/no value retain that setting in the next version. But then other values won't make sense and we get in the same pickle we have now with `use_textile` taking on three or more values, some of which don't even relate to Textile!

[...]

PR #920, opened 2017-07-16
popHelp

makss:
## This patch contains only `popHelp`, everything else will be taken to another patch.

---
- [x] first stage is a simple replacement of rpc - **Done**
- [x] second stage is Ajax to some element of design - **Done**
- [ ] Dashboard and custom help
- [x] plugin help
- [ ] plugin pophelp

---
## 1. pophelp

**pophelp** - Used {current-UI-language}_pophelp.xml  If there is no translation, then switching to the default language.

The XML structure has changed slightly for compatibility with the future format of the plugins help file.
```
<?xml version="1.0" encoding="utf-8"?>
<resources>
<help lang="en-gb">
    <group id="admin" title="Users panel">
	<item id="add_new_author" title="Add new author"><![CDATA[
		<h2>Users: Add new author</h2>
		<p>A password, welcome message and login instructions will be emailed to the new author.</p>
	]]></item>
	<item id="remember_login" title="Remember login" format="textile"><![CDATA[
h2. Remember login

When checked, you will remain logged in to your Textpattern account with the current web browser, until you choose to log out. If left unchecked, you will be automatically logged out after a period of inactivity.

Requires cookies to be enabled in your browser.
	]]></item>
    </group>

    <group id="prefs" title="Preferences panel">
	<item id="admin_side_plugins" title="Preferences: Use admin-side plugins?"><![CDATA[
		<h2>Preferences: Use admin-side plugins?</h2>
		<p>Admin-side plugins can change the operation of the administration interface. They might hook into already existing actions, like editing or publishing articles, or they might add new panels for added/extended functionality.</p>
		<p>If you would like to learn more about admin-side plugins, please see the <a rel="external" target="_blank" href="http://docs.textpattern.io/development/">Extending Textpattern <span class="ui-icon ui-icon-extlink">(opens an external link in a new window)</span></a> page in <a rel="external" target="_blank" href="http://docs.textpattern.io/">Textpattern User Documentation <span class="ui-icon ui-icon-extlink">(opens an external link in a new window)</span></a>. Plugins of all sorts are listed in the <a rel="external" target="_blank" href="http://textpattern.org/">Textpattern Resources <span class="ui-icon ui-icon-extlink">(opens an external link in a new window)</span></a> site.</p>
	]]></item>
	<item id="allow_article_php_scripting" title="Preferences: Allow PHP in articles?"><![CDATA[
		<h2>Preferences: Allow PHP in articles?</h2>
		<p>When enabled, this setting will allow use of PHP within articles. The author must have sufficient privileges to do so (by default, only <a rel="external" target="_blank" href="http://docs.textpattern.io/administration/user-roles-and-privileges">‘Publishers’ <span class="ui-icon ui-icon-extlink">(opens an external link in a new window)</span></a> and <a rel="external" target="_blank" href="http://docs.textpattern.io/administration/user-roles-and-privileges">‘Managing Editors’ <span class="ui-icon ui-icon-extlink">(opens an external link in a new window)</span></a> can).</p>
		<p>PHP code must have opening and closing PHP tags omitted, and also be enclosed within <a rel="external" target="_blank" href="http://docs.textpattern.io/tags/php"><code>&lt;txp:php&gt;</code> tags <span class="ui-icon ui-icon-extlink">(opens an external link in a new window)</span></a>.</p>
	]]></item>
    </group>
</help>
</resources>
```
---
## 2. Dashboard and custom help

* Dashboard. `?event=help&step=dashboard` or `?event=help`  (in developing)
* Allow display any help.xml file. `?event=help&step=custom`  (in developing)

---
### Expand the possibilities of plugin help:

* Multilanguage support in help files.
    * Detect the preferred language.
    * Select the language from the available
* Allow use `textile` and `html`
    * Allow mix `textile` and `html` blocks in one help file
* Separate unit for `css` (global for all languages)
* Separate unit for `javascript` (global for all languages)
* Support auto-build TOC based on XML tag `<group>`

### Minimal XML help

```
<?xml version="1.0" encoding="utf-8"?>
<resources>
<help lang="en-gb">
	<item format="textile"><![CDATA[
h3. What do you want to do next?
...
]]></item>
</help>
</resources>
```

### Structure of the XML help  (same as pophelp)
```
<?xml version="1.0" encoding="utf-8"?>
<resources>
<help lang="en-gb">
<group id="admin" title="Users panel">
	<item format="textile"><![CDATA[
h3. What do you want to do next?
...
]]></item>
	<item format="textile"><![CDATA[
h3. Other textile block, why not?
]]></item>
	<item><![CDATA[
<h3>Bulk HTML block</h3>
]]></item>
</group>
</help>

<help lang="uk-ua">
	<item format="textile"><![CDATA[
h3. Ukrainian help file
]]></item>
</help>

<css><![CDATA[
.qwerty {
	color: red;
}
]]></css>

<js><![CDATA[
	// js code here
//	alert('qwerty');
]]></js>
</resources>
```

### Example xml help file, install the plugin and go to the help section.

```
# Name: aks_help_test v0.1a (compressed)
# Type: Public plugin
# test xml help file
# Author: makss
# URL: http://makss.uaho.net/
# Recommended load order: 5

# .....................................................................
# This is a plugin for Textpattern CMS - http://textpattern.com/
# To install: textpattern > admin > plugins
# Paste the following text into the 'Install plugin' box:
# .....................................................................

H4sIAAAAAAAAA71V/W/iNhj+uf4r3kVIwNQkBO6AhZDqbpt2J3F3k9Zpk9oKmcQhKc5HbedS
VPV/32sntLSrtk7aZgmw34/H78djXup7Y/9O+m98q6A5sxbS9ya+RXdynTJerRWTSgunKKtV
Wgp9eOtbOVpIYz06aNa1yLRkPPatVKnKd11j5dQ0LZ2CKVdrZ771lQmZlYU+4bUjx6MGyPOt
mMlIZJXqtN7ct3QAcJtz0OFAknHW+UVlzDpA14VNnVedQu2rNg/E7qItRcxEJ3vbyRJOt/LY
zkMh5bxs1qnKuUnfWmx8r0Vtj2g1neKNwZkOqUtkaXnOyAJWYEhZsV1atUrsuXUWkkAwWdYi
YhL3JgFOtQEr7O3GCslJkCmWQ1KKnKol5nqrdIJh8M3F9z+8O393QdKJA7+lVEFcwr6soaGF
AlXqY4HWZ4RUgjmQ1EWkqwYVr7dZ0fZOsALTHvRiqugp9FLBkmW/PyR3BHD1ylrBEvr9RXvU
kUkUUCHofjBspVkCg55OdQkyyyvOcL/mJY3XUgnM9QBu/WK0v39a/chZzgplncLq43s8rz9/
MakMh3BHHMcxsPfknpBVJhVC+IR8Czc3+NU0DTEmV1dh4OrKhOSVJfqiUiagU8KGl9HuFJp0
D0Wpa7S5cYzy//0iLyRyFHeQTsL3Nd/Bh/NPqzbmwEXZ8xKgELuJCMcMqnd2TV/NoF93gmZF
RovHV9QR57B6h9otkVcNXH5mKqICCX553iouf6ZCIps6YuiFlzIapcgQjWqHJhIqoac3ut1w
tHptiIhv1HZIFTJoU+P7HgztsNUunroYwCWgXT5oQZ8aGHYecJHKXQ7953c/0N1ZPiRqh93m
PM1kh+4UT/HvGZfsL6C00598yOOO/G0bE2En4qGNz8ixQhMwFi0vSBBnXyHiVErs9G1lb0UW
o7O+7CWVHTHOLbhoslilpyAZ08+ByavwIcagCn9CSzCV9gK36tBchAv/DczxazEhzbYpx4/6
B+iT/yDiN88wSffzci8jKY+65tw0TKg90u8kKnkpfBAsXuCfnfHVtuhyfexxgrPrWoIeZvg2
BSOue0I5ggz6LVYfOW+8r/UMcR8Hih5GOKnz2Ay0CQ7dZDKbziffselo7M3no2g2nW5mo5jO
xtMk9pKxtbj/A5wTpLDvBwAA
```

---
---
## TODO

* plugin pophelp
* `?event=help&step=txp` - Short offline Textpattern help `{current-UI-language}_help.xml`
* `?event=help&step=custom` - Allow web developers to register and display their help file. For example, it will be convenient to leave the instruction to users.
* `?event=help&step=theme` vs `?event=theme&step=help`


philwareham:
@makss great work, I will test some UI methods as soon as I can to house the help system. I just need a few days getting our Textpattern new website launched in between my day job, so bear with me.

petecooper:
Tagging in https://github.com/textpattern/textpattern/issues/292 for completeness, seems to have sufficient overlap that it can be closed if (when) this gets merged.

makss:
`?event=help&step=custom&name=en-gb_pophelp` - show custom help file `en-gb_pophelp.xml` (testing rendering)

makss:
[Demo](http://makss.uaho.net/help-paged/) rendering `en-gb_pophelp.xml`. Javascript is not used.

philwareham:
Lovely stuff!

One thing I forgot to mention previously is the textfilter help (currently Textile only, but potentially other text filters too via plugins or (maybe) core)- I've left that out of the XML filer now. How should we handle help for those - any ideas?

Bloke:
Here's a comment [regarding textfilter help](https://github.com/textpattern/textpattern/issues/887#issuecomment-294174738). It's still undecided, but from 4.7.0, a registered Textfilter can only have a single help URL. For Textile, this is txstyle.org. For Markdown it could be daringfireball.com (or somewhere else depending on which version of Markdown the plugin author chooses).

For Leave Text Untouched and Line Breaks, if we deemed help necessary the links could be to `?event=help&step=popup&item=something` (??) which would thus reroute to our internal help system.

We also need to consider plugins here, btw. When a plugin is installed, the UI may require pophelp topics. So the API should be able to do this on lifecycle events:

a) install a language XML file in the lang directory.
b) update that file on upgrade.
c) delete that file when plugin deleted.

philwareham:
@bloke OK, that's good I think - reduces the work we have to do documenting textfilters.

How would plugins add their help to the system? Not sure I understand how that would work. It would be very handy to allow plugins to access/add to the help system though.

Regarding the pophelp symbols - I'd quite like to retire them in favour of a system roughly based on @makss demo (although limited to the help for the panel you are on) and what WordPress do. i.e. have a drawer or slide-out or something there all the help items for that panel are displayed in alphabetical order.

Getting rid of pophelps helps with things like preferences, where I think the system always add a pophelp icon. This is weird on plugins that add additional prefs currently as their is a broken pophelp link for every non-core preference.

We can keep the pophelp in a limited capacity if you need to provide a link off-site for textfilters and maybe some other things I can't think of right now.

Bloke:
+1 for a slide-out drawer and remove the inline 'i' links. Between the drawer/slider and the ability for people to add their own inline help on fields, the 'i' pophelp has had its day.

philwareham:
Getting rid of the 'i' like also helps keyboard navigation through the system (as you have to tab past each pophelp link currently which is super-annoying).

Bloke:
Good point about the keyboard nav. that is annoying to double-tab to the next field (especially during setup).

We'll need to look at `inputLabel()` of course. One of its options is combined pophelp/inline help so would probably need refactoring to ditch the pophelp component in some backwards-compatible way.

makss:
> It's still undecided, but from 4.7.0, a registered Textfilter can only have a single help URL.

If we use the help output in some UI element, then we can return some html string:
`Textile help <a href='http://txstyle.org'>txstyle.org</a>`

> We also need to consider plugins here, btw. When a plugin is installed, the UI may require pophelp topics. So the API should be able to do this on lifecycle events:
>
>a) install a language XML file in the lang directory.
b) update that file on upgrade.
c) delete that file when plugin deleted.

Why make it so complicated? :-)
Can be solved plugin-pophelp using prefixes. `abc_plugin-some_my_pophelp` or `abc_plugin@some_my_pophelp` or somehow else - Indicates that the pophelp needs to be taken from the `html` column in the plugin table. We will have to add some wrapper to the pophelp, for example:

```
<help lang="en-gb">
	... Ordinary help ...
</help>
<pophelp lang="en-gb">
	<item id="add_new_author" title="Add new author" />
	<item id="some" title="Other item" />
</pophelp>
```
Override pophelp core calls: `register_pophelp('core_pophelp_id', 'plugin_pophelp_id')`
Example: `register_pophelp('add_new_author', 'abc_plugin@my_new_author')`

Bloke:
Interesting idea about using the existing `html` column to hold pophelp too. That makes a lot of sense, and also permits multi-lingual content. I was just assuming that our pophelp API would be reading only from file, but you're right: it makse sense to also allow it to read from a DB field.

In which case, we need to increase the size of the html field to MEDIUMTEXT and figure out a way to make the `html` field backwards-compatible. Perhaps if it looks like an XML document (i.e can be parsed as one) then we treat it like that, otherwise we assume it's an old-school Textile document. Advantages: the XML attributes determine what Textfilter to apply, if any. Default: html? Textile?

I like this idea. So let's firstly define the help column schema. Do we need a full wrapper round everything or can it be automatically assumed?

``` xml
<docs>
<help lang="en-gb" format="textile">
	... Ordinary plugin help in Textile format...
</help>
<help lang="de-de">
	... German plugin help in HTML format...
</help>
<pophelp lang="en-gb">
	<item id="smd_option_1" title="Option 1">Does cool stuff.</item>
	<item id="smd_some_thing" title="Cool option" format="textile">
This option allows you to:

* Blah
* Blah blah
* Blob

It also washes your car for you.
        </item>
	<item id="and_this" format="textile">
h2. Another option

Select this to pet a kangaroo.
        </item>
</pophelp>
<pophelp lang="de-de">
        <item id="smd_option_1" title="Option 1">Description of cool stuff in German.</item>
        ...
</pophelp>
</docs>
```

My only slight concern is that it becomes slightly more difficult to write. I'm sure Plugin Composer could help out and bundle all the stuff up for you, but if you're doing it by hand it might be a chore. But then the benefits to users are immense, so it's worth it.

makss:
@Bloke, In the previous post I greatly reduced the xml example, I fixed the post above. I think to make a 100% compatible format for `help` and `pophelp`. (Tag `<group>` and attribute `id`, `title` in `<item>` tag - optional)

Btw, can not allocate `pophelp` separately, but pull them out from `help` tag  by `id`  - `<item id='smd_option_1'>`. In this case, we will be able to view the help as a whole and pull out individual items as `pophelp` prompts.

philwareham:
Right, I'm ready to look further into this. What do you need from me? A way of housing the help within our UI, yes?

makss:
@philwareham, Sorry for the late reply, I'm back online again. To rest well. :)

Yes, I need help with UI. We need to create some html element(`div`) and with ajax we'll need to draw the pophelp into it. Also need to make it visible, and when closed, invisible.

Example ajax request GET or POST
`?event=help&step=pophelp&app_mode=async&item=sitename`

makss:
The first attempt ajax pophelp.

makss:
It seems the ajax pophelp is already working. I ask you to test it.

* I made the caching of ajax requests to the attribute of the link `data-item`.
* For unauthorized users and during installation, the inclusion of a pophelp in the page itself is used (attribute `data-item`).

---

Thoughts that came only afterwards:
* Can include pophelp in the page itself and do not use ajax.
* It is possible to convert `en-gb_pophelp.xml` to `en-gb_pophelp-json.js`, to connect it to the page as javascript. Its size will be approximately 100Kb, the browser will cache it once and will no longer request it. In this case, too, no ajax requests are needed.

philwareham:
Just a note that I haven't forgotten about this work - just been tied up with other items first. Once theme branch is merged we can give this PR some time and effort.

philwareham:
Hi all,

We (Stef, I and Jakob) discussed the help system at length when we met up on Saturday. After some thought we felt that the best approach *for now* is to have the modal popups (pophelps) - as you have done in this PR - and park the drawer UI concept containing the whole page help list for now.

The reasons for this:

1. The individual pophelps are directly accessible from the item. The overall page help listed alphabetically would not be so direct - you'd have to scroll through a lot of help text to find the specific item's help (especially on the prefs panels).
2. The overall page help system I can't find a UI for that I'm totally happy with using, and that works well across desktop, tablet and mobile.
3. We want to get Textpattern 4.7 out before Christmas and the extra development time to get overall page help into core would likely delay that release.

However, I will keep thinking of ways to improve the help system and I do generally like the overall page help - so not saying it'll never work. Just not right now.

And in the here and now I have the following observations that I hope can improve the current pophelps further...

1. We could have a pref in the core that allows admin to turn off pophelps completely if they are comfortable with the system and find the help links intrusive. This could simply be an extra `class` added to the `<body>` when the pref is turned on. Then used like so:

    .no-inline-help .pophelp {
        display: none;
    }

That takes pophelps out of the tab chain.

2. Currently all prefs on prefs panel force a pophelp link, including prefs added by plugins, even when pophelp doesn't exist for that help item. We need to allow prefs to be set with pophelps or not.

Cheers!

makss:
* Added pref `enable_admin_pophelp` (You need to logout and login to have this auto option created)
* Prefs panel: show only existing pophelp

I still need to carefully check this patch, remove unnecessary, once again check the format of xml.

PR #930, opened 2017-09-02
Switch gTxt `invalid_expirydate` to `invalid_expiredate`

petecooper:
Uniformity: gTxt strings refer to 'expire' (multiple mentions), not 'expiry' (zero mentions).

There's a request for addition in Textpacks (https://github.com/textpattern/textpacks/issues/217), but if we can change this out then the more appropriate gTxt (`invalid_expiredate`) can be added to be a better fit.

(Please and thank you.)


philwareham:
Agreed, thanks Pete!

PR #932, opened 2017-09-13
Adds Skin package (includes txp_skin.php changes)

NicolasGraph:
As discussed with @Bloke  and @philwareham by email, here is a pull request introducing an initial Skin package to manage related features. Basics work, advanced features and callbacks still need some tests. Happy debug!

#### Main classes
- [Main](https://github.com/NicolasGraph/textpattern/blob/themes/textpattern/vendors/Textpattern/Skin/Main.php) manages multiple themes features;
- [Skin](https://github.com/NicolasGraph/textpattern/blob/themes/textpattern/vendors/Textpattern/Skin/Skin.php) manages a single theme and its assets (pages, forms and styles);
- [Pages](https://github.com/NicolasGraph/textpattern/blob/themes/textpattern/vendors/Textpattern/Skin/Pages.php) manages a defined skin related pages;
- [Forms](https://github.com/NicolasGraph/textpattern/blob/themes/textpattern/vendors/Textpattern/Skin/Forms.php) manages a defined skin related forms;
- [Styles](https://github.com/NicolasGraph/textpattern/blob/themes/textpattern/vendors/Textpattern/Skin/Styles.php) manages a defined skin related styles.

#### Admin features
- create a new theme with default pages, forms and styles* (was part of txp_skin.php);
- edit an installed theme (was part of txp_skin.php);
- Import/update a theme from flat files;
- duplicate an installed theme;
- export/backup an installed theme as flat files;
- Delete an installed theme which is not in use.

While the Import feature code was inspired by [rah_flat](https://github.com/gocom/rah_flat), it includes changes such as introducing `setMaxDepth` to the `recursiveIteratorIterator` to avoid useless iterations (mainly because of the `composer.json` file introduction) or grouping db queries for better performances.

The skin tab includes the features above for whole themes but the package is ready for defined theme parts manipluation.

[*] _The [`$essential`](https://github.com/NicolasGraph/textpattern/blob/themes/textpattern/vendors/Textpattern/Skin/Forms.php#L42) property of the `Forms` class (also available for `Pages` and `Styles` classes) should be able to replace the [`get_essential_forms()`](https://github.com/textpattern/textpattern/blob/themes/textpattern/lib/txplib_misc.php#L2909) function._ 

#### Theme/skin directory

```
|-- abc_theme
|  +-- composer.json
|  +-- pages
|  |  +-- default.txp
|  |  +-- …
|  +-- forms
|  |  +-- article
|  |  |  +-- default.txp
|  |  |  +-- …
|  |  +-- comment
|  |  +-- …
|  +-- styles
|  |  +-- default.css
|  |  +-- …
```

The theme informations are managed from the `composer.json` file.
Pages and forms accepts the following extensions: `.txp` and `.html`.

#### Misc.

- I changed the theme related event+step couples by editing steps to have something like `skin.import` instead of `skin.skin_import`. While it seems more logical to me I'm not sure about the consistency; other files seem to also include the event name in step names…
- You will find a basic textpack [here](https://github.com/NicolasGraph/textpattern/blob/themes/textpattern/vendors/Textpattern/Skin/gb.textpack).

### DIFF
--- a/textpattern/include/txp_skin.php
+++ b/textpattern/include/txp_skin.php
@@ -29,6 +29,7 @@
  */
 
 use Textpattern\Search\Filter;
+use Textpattern\Skin\Main as Skins;
 
 if (!defined('txpinterface')) {
     die('txpinterface is undefined.');
@@ -37,20 +38,17 @@ if (!defined('txpinterface')) {
 if ($event == 'skin') {
     require_privs('skin');
 
-    global $all_skins;
-    $all_skins = safe_column('name', 'txp_skin', "1=1");
-
     $available_steps = array(
-        'skin_change_pageby' => true,
-        'skin_list'          => false,
-        'skin_delete'        => true,
-        'skin_save'          => true,
-        'skin_edit'          => false,
-        'skin_multi_edit'    => true,
+        'change_pageby' => true,
+        'list'          => false,
+        'edit'          => false,
+        'save'          => true,
+        'import'        => false,
+        'multi_edit'    => true,
     );
 
     if ($step && bouncer($step, $available_steps)) {
-        $step();
+        call_user_func('skin_'.$step);
     } else {
         skin_list();
     }
@@ -79,9 +77,18 @@ function skin_list($message = '')
     if ($sort === '') {
         $sort = get_pref('skin_sort_column', 'name');
     } else {
-        if (!in_array($sort, array('name', 'title', 'version', 'author', 'section_count', 'page_count', 'form_count', 'css_count'))) {
-            $sort = 'name';
-        }
+        $sort_options = array(
+            'title',
+            'version',
+            'author',
+            'section_count',
+            'page_count',
+            'form_count',
+            'css_count',
+            'name',
+        );
+
+        in_array($sort, $sort_options) ? $sort = 'name' : '';
 
         set_pref('skin_sort_column', $sort, 'skin', 2, '', 0, PREF_PRIVATE);
     }
@@ -93,37 +100,11 @@ function skin_list($message = '')
         set_pref('skin_sort_dir', $dir, 'skin', 2, '', 0, PREF_PRIVATE);
     }
 
-    switch ($sort) {
-        case 'title':
-            $sort_sql = 'title '.$dir;
-            break;
-        case 'version':
-            $sort_sql = 'version '.$dir;
-            break;
-        case 'author':
-            $sort_sql = 'author '.$dir;
-            break;
-        case 'section_count':
-            $sort_sql = 'section_count '.$dir;
-            break;
-        case 'page_count':
-            $sort_sql = 'page_count '.$dir;
-            break;
-        case 'form_count':
-            $sort_sql = 'form_count '.$dir;
-            break;
-        case 'css_count':
-            $sort_sql = 'css_count '.$dir;
-            break;
-        case 'name':
-        default:
-            $sort_sql = 'name '.$dir;
-            break;
-    }
-
+    $sort_sql = $sort.' '.$dir;
     $switch_dir = ($dir == 'desc') ? 'asc' : 'desc';
 
-    $search = new Filter($event,
+    $search = new Filter(
+        $event,
         array(
             'name' => array(
                 'column' => 'txp_skin.name',
@@ -142,22 +123,21 @@ function skin_list($message = '')
 
     list($criteria, $crit, $search_method) = $search->getFilter();
 
-    $search_render_options = array(
-        'placeholder' => 'search_skins',
-    );
-
+    $search_render_options = array('placeholder' => 'search_skins');
     $total = safe_count('txp_skin', $criteria);
 
     echo n.'<div class="txp-layout">'.
         n.tag(
             hed(gTxt('tab_skins').popHelp('skin_category'), 1, array('class' => 'txp-heading')),
-            'div', array('class' => 'txp-layout-4col-alt')
+            'div',
+            array('class' => 'txp-layout-4col-alt')
         );
 
     $searchBlock =
         n.tag(
             $search->renderForm('skin', $search_render_options),
-            'div', array(
+            'div',
+            array(
                 'class' => 'txp-layout-4col-3span',
                 'id'    => $event.'_control',
             )
@@ -168,8 +148,10 @@ function skin_list($message = '')
     if (has_privs('skin.edit')) {
         $createBlock[] =
             n.tag(
-                sLink('skin', 'skin_edit', gTxt('create_skin'), 'txp-button')
-                , 'div', array('class' => 'txp-control-panel')
+                sLink('skin', 'edit', gTxt('create_skin'), 'txp-button').
+                Skins::renderImportForm(),
+                'div',
+                array('class' => 'txp-control-panel')
             );
     }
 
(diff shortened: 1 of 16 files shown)


philwareham:
Brilliant work! We can give this a bit of testing for any serious bugs and then merge and fine tune in dev branch from there on.

philwareham:
@Bloke can you check through the latest `themes` branch and make any tweaks you deem necessary before it enters `dev` branch? I also think we need to move the default theme from `setup` directory to `themes` directory and make sure that works during installation, but if you think that can/should be done after `dev` merge then fine. I'd need to change the structure of the default theme anyway as forms will need to go into subdirectories instead of having the form type in the filename I believe.

Bloke:
Sure. I had a play this morning and seems good. Had some warnings importing a theme, but I think that was down to file permissions on my server so when I imported the theme second time, it complained.

Yes, the default theme needs moving so that we always have a 'default' entry that the theme manager can read from/write to.

On that note, if we rename a theme in the UI, shouldn't it rename the corresponding subdir in the 'themes' directory? Otherwise, how will a plugin (or core in future) be able to manage themes for export if it doesn't have maintain a coherent link between DB and file system?

NicolasGraph:
> On that note, if we rename a theme in the UI, shouldn't it rename the corresponding subdir in the 'themes' directory?

It should probably; I missed that case.

NicolasGraph:
@Bloke, [this](https://github.com/textpattern/textpattern/commit/06952e93003ee4e827ccb214113641c4c189a501) seems to work, do I need to send a new pull request?

Bloke:
>  if no new theme is found, the import form is replaced with the following message: "No new theme to import"

Yes, I saw that. That message would also appear if we moved the default theme to the themes directory, because there would be "no new themes" - default is always loaded. So I think it's safe to move the default theme here.

Just thinking out loud, is the 'no new theme' message even necessary? If there's nothing that can be done, should we just omit the block altogether (or rather, render `''`, but still raise a pluggable_ui callback so that plugins can do their own thing if they wish)?

Couple of minor niggles if I stick the site in debugging mode:

When deleting a theme:
```
Warning "array_splice() expects parameter 2 to be integer, string given"
in textpattern/vendors/Textpattern/Skin/Skin.php at line 586.
```

When exporting a theme:
```
Notice "Undefined index: authors"
in textpattern/vendors/Textpattern/Skin/Skin.php at line 536
```

> do I need to send a new pull request?

Since the old PR has been merged, then yes I think a new one is required. But you can roll a bunch of changes/fixes into it if you like.

NicolasGraph:
> Just thinking out loud, is the 'no new theme' message even necessary?

It was just a way to mention that themes import is possible.

> Couple of minor niggles if I stick the site in debugging mode […]

Ouch; I turned the production status to see how fast it was in "Live" mode and forgot to turn it back. I'll take look at these bugs. Thanks for the report!

Bloke:
Another little bug: it's not possible to Save the theme details.

STR:

* Visit Themes panel
* Click to edit any theme's info (edit step).
* Make no changes and click Save.

AR:
Error issued: duplicted_skin.

ER:
Record is updated.

If `old_name` matches `name`, an update should occur. Only if `old_name` and `name` are different AND `name` is the same as any other theme name, should the error be triggered.

Side note: typo in the word 'duplicated_skin' (missing 'a').

Bloke:
> It was just a way to mention that themes import is possible.

Okay, that's fine with me. It does advertise the fact that theme support is available so, yeah, lets' keep it.

phiw13:
(not to distract from the discussion)

I’d love to take this for a spin. However, how do I get past this: https://github.com/textpattern/textpattern/compare/themes#commitcomment-23012379

(I’m stuck on the diagnostics panel with the message:

```
User_Error "Row size too large. The maximum row size for the used table type, not counting BLOBs, is 65535. This includes storage overhead, check the manual. You have to change some columns to TEXT or BLOBs"
in /Users/username/Sites/_txp47/textpattern/lib/txplib_db.php at line 409.
```

yes, MySQL is utf8mb4…)

Bloke:
@phiw13 Not sure without installing this on an empty 4.6.2 and letting it do the upgrade. I'll try it later. Index sizes or column types probably need tweaking, sorry.

NicolasGraph:
I also found a bug on export which seem to appears for themes created in the UI only. I'll work on it.

philwareham:
Any specific reason `composer.json` has been chosen to store theme metadata? That filename is kind of exclusively reserved for Composer installations. You should probably use `manifest.json` here instead.

If we do eventually allow theme installations via Composer (something @gocom could probably add to his textpattern-installer if he so desired) then we would need a proper `composer.json` file for that use. [See here default theme changed to support this themes branch](https://github.com/textpattern/textpattern-default-theme) but with a `manifest.json` file in the theme itself (look in the `dist` directory).

NicolasGraph:
> I also found a bug on export which seem to appears for themes created in the UI only. I'll work on it.

In fact, it is not really a bug but kind of… If a page, form, or style as no content, the file is created on export but an error message is displayed. Not sure what should be done. Empty templates should maybe not be exported at all?

> Any specific reason composer.json has been chosen to store theme metadata?

@bloke ask me to work with this file to be able to add a composer support without any extra file I think.

philwareham:
The composer.json file only works in the root of a project, [see docs](https://packagist.org) and contains details expected by Composer.

You'd need a project structure like this on GitHub:

````
+-- composer.json
|-- dist
|  |-- abc_theme
|  |  +-- manifest.json
|  |  +-- pages
|  |  |  +-- default.txp
|  |  |  +-- …
|  |  +-- forms
|  ||  |  +-- article
|  |  |  |  +-- default.txp
|  |  |  |  +-- …
|  |  |  +-- comment
|  |  |  +-- …
|  |  +-- styles
|  |  |  +-- default.css
|  |  |  +-- …
|-- src
|  |  +-- …
````

From my understanding, when once the project is registered on Packagist, you could feasibly go to the root directory of your site and run `composer require abc/abc_theme` and (if composer.json is configured right) the theme `abc_theme` would be installed from `dist` into the `themes` directory.

TLDR; please can the file be `manifest.json` instead of `composer.json`.

Also, because the composer.json file ([example here](https://github.com/textpattern/textpattern-default-theme/blob/master/composer.json)) needs to know the target directory to install into, we would need to ensure `themes` is not changeable to something else. *Or* I guess that is something that could be overcome by textpattern-installer if that route was available instead (it could remap to a vendors directory and bypass `themes`. Dunno, just thinking out loud.

Bloke:
Phil's right. I was going to mention that it should be `manifest.json`. If I mentioned `composer.json`, then I'm sorry: must have been a bad day!

Bloke:
> Empty templates should maybe not be exported at all?

Part of me thinks they should be exported, if possible, without error. Some templates might have 'optional' elements that a designer leaves as a placeholder for people to fill in if they want a particular feature. But then, equally I suppose, an empty file isn't much use and just takes up space. Not sure.

Options:

1. Theme designer writes in the instructions "If you want to use this feature, create a new Form called blah-blah and populate it with content". OR:
2. Theme designer writes in the instructions "If you want to use this feature, navigate to the empty Form called blah-blah and populate it with content".

The first one means the onus is on the user to create the form with exactly the correct name, the second is down to the designer. I'm leaning towards the designer right now, for ease of use. Especially given that we have "essential" forms that a theme might not use, depending on tag usage in the theme. To avoid confusion, the designer may empty that content and thus save an empty 'required' file in the theme structure. That's perfectly valid and we wouldn't want to issue an error in thise case, so perhaps we should treat other "empty" content likewise and permit them to be stored as empty files?

NicolasGraph:
> Phil's right. I was going to mention that it should be manifest.json. If I mentioned composer.json, then I'm sorry: must have been a bad day!

Ok, no worry, it will be manifest.json. Not sure now if it was you or me who were/was in a bad day…

> Part of me thinks they should be exported, if possible, without error.

Thanks, it is also what I thought this the weekend.
I didn't test empty files, I'm going to do it…

NicolasGraph:
> But then, equally I suppose, an empty file isn't much use and just takes up space.

Empty files are imported correctly as far as I tried.

NicolasGraph:
@Bloke, I get a bug on skin update coming from the SQL query. Did you consider to create a unique index from the skin + name couples? It would facilitate the fix in the query using `ON DUPLICATE KEY`.

NicolasGraph:
Sorry @Bloke, indexes were not the problem in the above commit. I was!

Issue #933, opened 2017-09-14
The readable output of json_encode()


makss:
We use JSON for ajax queries, preferences and in the `txp_escape` tag.

The encoded UTF-8 JSON looks like this:

> textpattern.textarray = {"are_you_sure":"\u0412\u0438 \u0432\u043f\u0435\u0432\u043d\u0435\u043d\u0456?","cookies_must_be_enabled":"Cookies \u043f\u043e\u0432\u0438\u043d\u043d\u0456 \u0431\u0443\u0442\u0438 \u0432\u043a\u043b\u044e\u0447\u0435\u043d\u0456, \u0449\u043e\u0431\u0438 \u0412\u0438 \u043c\u043e\u0433\u043b\u0438 \u043a\u043e\u0440\u0438\u0441\u0442\u0443\u0432\u0430\u0442\u0438\u0441\u044c Textpattern.","form_submission_error":"\u041d\u0430 \u0436\u0430\u043b\u044c, \u0444\u043e\u0440\u043c\u0430 \u043d\u0435 \u043c\u043e\u0436\u0435 \u0431\u0443\u0442\u0438 \u0437\u0431\u0435\u0440\u0435\u0436\u0435\u043d\u0430. \u0411\u0443\u0434\u044c \u043b\u0430\u0441\u043a\u0430, \u043f\u043e\u0432\u0442\u043e\u0440\u0456\u0442\u044c \u0441\u043f\u0440\u043e\u0431\u0443 \u043f\u0456\u0437\u043d\u0456\u0448\u0435.","list_options":"list_options","ok":"ok","publish":"\u041e\u043f\u0443\u0431\u043b\u0456\u043a\u0443\u0432\u0430\u0442\u0438","save":"\u0417\u0431\u0435\u0440\u0435\u0433\u0442\u0438","toggle_all_selected":"\u0412\u0438\u0431\u0440\u0430\u0442\u0438 \u0432\u0441\u0435","help":"\u0414\u043e\u043f\u043e\u043c\u043e\u0433\u0430"}

I propose to add a check `JSON_UNESCAPED_UNICODE` constant and use it, if it is available.
Output:

[...]

PR #934, opened 2017-09-14
The readable output of json_encode()

makss:
Issue #933


makss:
Here the situation is the same as with plain text, UTF-8 text remains as is, only quotes and other service characters are escaped. If there is a problem with the collation, it will be common for both plain text and json_encode().

bloatware:
That's true, but one might use, say, English on a non-UTF-8 collation without any problem, and then install a plugin in Ukrainian. Wouldn't saving its preferences be problematic without encoding? An edge case, sure.

bloatware:
Ok, let's test in wild. Thanks @makss !

PR #936, opened 2017-09-15
Include jquery-ui.js in the Setup

makss:
Fix JS warning:
```
jQuery.Deferred exception: this.dialog is not a function jQuery.fn.txpDialog
TypeError: this.dialog is not a function
```


Bloke:
N.B.: Reverts db6bffca

philwareham:
OK, my bad. Didn't think jQuery UI link was needed in setup.

PR #941, opened 2017-09-29
Universal setup for dev/themes branch

makss:
It seems it works.  ;-)
Please test.

**In `themes` branch:**
I do not understand why, but to import the class `Skin->import()` requires the right to write to the directory?!  Therefore, you need to either give the right to write to the directory of the skins or fix `Skin` class.

btw, now `dev` branch allow select and install public themes during setup.


philwareham:
Thanks makss. I've merged so we can test in-situ.

philwareham:
I've also pushed these changes to the `themes` branch for testing. We need to note that the files in setup *won't* be there once the themes branch is ready. Themes (including the default theme that ships with the installation) will only be installed from there - not from both locations.

In the theme branch I've removed the theme from setup directory now - it causes some issues currently when installing but I'm sure it's all fixable.

Issue #839, opened 2016-11-09
Use JSON file to check for new releases (instead of RPC server)


philwareham:
The new package.json file (that will reside in both GitHub repo and in root of Textpattern.com) will be used to check if new versions are available, instead of the old RPC server (which we are gradually moving away from).

We propose checking the Textpattern.com root first, then if unavailable check GitHub repo as a fallback.

rwetzlmayr:
How will a maintainer deploy textpattern.com/package.json?

I suggest to take a look at the current implementation, i.e. Kusor's `txp_current_version` plugin the at the RPC server as it allows maintaining this information from the Textpattern admin area without needing any other server access methods (ssh et al.)

philwareham:
They would need FTP access to .com to update the JSON file, I guess. Other option is to serve the JSON just from GitHub but:

a) That relies on a third party.
b) It may break GitHubs terms of use

If the `txp_current_version` plugin is something simple that can run on .com without modifications to Textpattern core, then I'm interested. It would need some development so that we can provide a stable release number, as well as separate prerelease number when beta versions are available. The diagnostic panel can then give people the option of using a beta when available, or stick to stable releases.

rwetzlmayr:
> If the txp_current_version plugin is something simple that can run on .com without modifications to Textpattern core

It is. You can look at the code on `rpc.textpattern.com/textpattern/index.php?event=plugin&step=plugin_edit&name=txp_current_version`.

philwareham:
OK, great, I'll investigate.

Basically the major reason for having the GitHub fallback is if the .com domain ever disappears (which we all know is a possibility due to ownership). So even if we use the plugin on .com, I'd still like to fallback to GitHub. Otherwise in the (hopefully unlikely) event that .com fails - users could still be notified of an upgrade.

rwetzlmayr:
I think and hope that the actual discovery algorithm (how a client server finds their `package.json` on the web) is not related to the method used to generate this resource on textpattern.com (or github.com, or wherever).

philwareham:
Not wishing to pass the buck, but @Bloke will be wrong on this, so he is best placed to answer your concerns. I just opened the issue based on discussions we had at the weekend.

rwetzlmayr:
I just discussed the issue based on it being open ;)

petecooper:
Aside: a Textpattern page+form or page+form+article combination on textpattern.com would achieve the same effect as a proper file.

philwareham:
[...]

Issue #871, opened 2017-01-03
UX: managing expectations for Textpattern upgrades from 4.7


petecooper:
Scenario: RPC is not available, and a version check fails. The `problem_connecting_rpc_server` gTxt is displayed, and the `textpattern_version_update` pophelp is linked alongside, along with instructions on upgrading.

New pophelp or reuse an existing one? Rough screenshot attached, please note I'm on a new laptop and don't have screen grabbing software with cursors installed yet -- the URL is what's shown when hovering on the pophelp thing (blue arrow for…erm…clarity).

![rpc](https://cloud.githubusercontent.com/assets/226578/21613845/e64fdcb4-d1ce-11e6-8e93-e204be028dc2.png)

Bloke:
This is an interesting one. With reliance on RPC disappearing in 4.7.0, we'll be looking at getting latest available stable version number from (potentially) two places: a text file on textpattern.com, and a file on GitHub. This makes it far less likely that we'll see any such error, unless both services are down or there's no connection to the outside world.

In which case, I wonder if it's worth going a step further now: if the admin side can't connect to an external resource for version checking, so what? Do we need to inform anyone? Chances are that next time their site tries (there's a 24 hour throttle) it'll be up, unless the admin side is in an intranet situation or used for local testing.

Whaddya think? Is there any point complaining if external resources like .com and GitHub aren't available?

philwareham:
No, I don't think we need to complain in this case.

petecooper:
Complain? No. I agree with @philwareham. Inform? Via diagnostics in high mode, possibly.

I'd counter the headline with some form of diagnostics (high or low) that indicated the most recent check for an update, on the main source (.com or GitHub) and secondary source (opposite of main) if main timed out. That's outside the scope of this issue, and a low priority discussion. That provides some kind of feedback to say there a check is taking place, without explicitly mentioning it in diagnostics.

It'd also be rather helpful if some kind of top-level indicator on the nav bar or footer could be set when an update is available. Take the following scenario:

* Textpattern site runs just fine and dandy, thanks
* Update check happens, new version available
* Diagnostics panel never gets checked, because Textpattern site runs just fine and dandy, thanks.

I'm hesitant to push this for discussion because it only came about when RPC went AWOL for a short time, and then came back, because DNS. Shrug.

[...]

PR #942, opened 2017-09-29
Use JSON file to check for new releases

makss:
Issue #839 

TODO: Need change textpack string `problem_connecting_rpc_server`

### DIFF
--- a/textpattern/include/txp_diag.php
+++ b/textpattern/include/txp_diag.php
@@ -651,13 +651,6 @@ function doDiagnostics()
 /**
  * Checks for Textpattern updates.
  *
- * This function uses XML-RPC to do an active remote connection to
- * rpc.textpattern.com. Created connections are not cached, scheduled or
- * delayed, and each subsequent call to the function creates a new connection.
- *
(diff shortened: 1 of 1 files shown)


philwareham:
Nice! And fast work too! The only thing I’m thinking is that in the (albeit unlikely) event Textpattern.com domain is lost, it’d be better to use the .io domain here. It points to the same server and file anyway but we own that domain as a team.

makss:
Fixed ;-)

philwareham:
brillisnt, thanks! I’ll test this with a fake version update soon and then we can close the issue. I’ll also note the text pack string change.

Bloke:
And just for completeness: https://github.com/textpattern/textpattern/issues/871#issuecomment-271316405. Although that request has been sidelined (for now), it'd be good if we could expose the date of the last time a version-check attempt was made (in high diagnostics). But if the check fails during the attempt, for whatever reason, just silently fail.

makss:
btw, it seems we had a broken `last_update_check`, now I will fix it.

PR #944, opened 2017-09-29
Show `last_update_check` in detailed diagnostics

makss:
Issue #942


Bloke:
@makss Does this change still work for people who upgrade from older versions that have a serialized array stored in prefs?

makss:
Yes, it will work.

makss:
btw, when upgrading, this value is [deleted](https://github.com/textpattern/textpattern/blob/adec4dcf26eddb3f9953a04da2041ce44e7bfbb0/textpattern/update/_update.php#L78)

Bloke:
@makss: Ah yes, I'd forgotten about deleting that pref on upgrade. Brilliant, your method is much neater, thank you.

PR #959, opened 2017-10-23
Fixes error status

NicolasGraph:
The error status was not the right one on multi-edit depending of the order of the skin throwing the error.


philwareham:
Thanks!

PR #961, opened 2017-10-24
Themes branch: JSON pretty print for PHP 5.3.x

NicolasGraph:
Uses 4 spaces for indentations.

### DIFF
--- a/textpattern/vendors/Textpattern/Skin/Skin.php
+++ b/textpattern/vendors/Textpattern/Skin/Skin.php
@@ -571,8 +571,80 @@ namespace Textpattern\Skin {
         {
             return (bool) file_put_contents(
                 $this->getPath(static::$file),
-                json_encode($contents, JSON_PRETTY_PRINT)
+                $this->JSONPrettyPrint(json_encode($contents)) // PHP 5.4+ => json_encode($contents, JSON_PRETTY_PRINT)
             );
+        }
+
+        /**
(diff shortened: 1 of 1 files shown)


makss:
There is also [JSON_UNESCAPED_UNICODE](https://github.com/textpattern/textpattern/issues/933), this is more unpleasant thing than formatting.

makss:
There is also [JSON_UNESCAPED_UNICODE](https://github.com/textpattern/textpattern/issues/933), this is more unpleasant thing than formatting.

NicolasGraph:
Thanks @makss.

philwareham:
Do you think we need a `JSON_UNESCAPED_SLASHES` setting too - JSON doesn't require escaped forward slashes. At the moment when you export a theme the `manifest.json` contains for example:

    "author_uri": "https:\/\/example.com\/"

When JSON is able to display this validly:

    "author_uri": "https://example.com/"

Not a problem, just looks cleaner without escaped forward slashes.

makss:
@philwareham , Good point!
Perhaps it's worth adding this to the entire Textpattern.
[Line](https://github.com/textpattern/textpattern/blob/e09bf9313e430fdf645cf1cae33425b3f36a8606/textpattern/lib/constants.php#L938) 
```
define('TEXTPATTERN_JSON', JSON_UNESCAPED_UNICODE);
```
change to
```
define('TEXTPATTERN_JSON', JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES);
```

A little later I'll check it out.

PR #963, opened 2017-10-25
Themes branch: rework (fixes #945)

### LINKED ISSUE #945: Themes branch: duplicating a duplicate results in index error
### Expected behaviour

When duplicating a theme, the duplicate should be made without error.

### Actual behaviour

The duplicate process throws a `Duplicate entry '{some-page-name}' for key 'name_skin'`.

### Steps to reproduce

* Export the default skin as a copy using the multi-edit tool.
* Note the name of the skin is appended with the date + time of export. This is good.
* Select the checkbox of this exported skin in the Themes list.
* Choose 'Duplicate' from the multi-edit tool and confirm the action.

The problem is due to the limited key length in use on the `name_skin` index. @NicolasGraph what do you think we should do? Increasing the index size is possible but at some point it'll be too large and cause errors. Where would we impose the limit? Other option is to _prepend_ the theme name with something different or put the date+time first, but that's not as friendly.

NicolasGraph:
Working on https://github.com/textpattern/textpattern/issues/945, I thought it would be useful to amend skin related classes/methods to stick more to the related processes available in the admin […].

```
Work in progress…
```

### Multiple themes edit

Use the `Skins` class.

#### Main  methods

- update (re-import);
- duplicate;
- export;
- delete.

As for now each method return `false` on error.

##### Export themes

```php
Txp::get('\Textpattern\Skin\Skins', array('first_theme', 'second_theme))->export()
```

or, rendering a message on error:

```php
$Skins = Txp::get('\Textpattern\Skin\Skins', array('first_theme', 'second_theme));
($Skins->export() === false) ? call_user_func_array('trigger_error', $Skins->getResults()) : '';
```

### Single theme edit

Use the `Skin` class.

### Main additional  methods

- create;
- import;
- edit;
- duplicateAs.

As for now each method return `false` on error.

##### Create a theme

```php
$new = array(
    name        => 'new_theme',
    title       => 'New theme',
    version     => '1.0.0',
    description => 'A brand new theme',
    author      => 'Me',
    author_uri  => 'http://my-site.com'
);

Txp::get('\Textpattern\Skin\Skin')->create($new);
```

Here the skin property is not set on instanciation as the skin does not exist yet in the DB or the _themes_ directory. To edit an existing theme with the same `$new` related infos, we would proceed like so:

```php
Txp::get('\Textpattern\Skin\Skin', 'my_theme')->edit($new);
```

##### Import a theme

```php
Txp::get('\Textpattern\Skin\Skin', 'my_theme')->import()
```

### multiple assets edit

Use the `Pages`, `Forms` or `Styles` class.

[…]

```
Work in progress…
```


NicolasGraph:
This PR can be merged, it fixes https://github.com/textpattern/textpattern/issues/945 and adds important code changes. There will probably be few more things, including callback changes according to what will be decided in https://github.com/textpattern/textpattern/issues/964 but main features are ready. Feedbacks welcome; have a good weekend.

Bloke:
Thanks for this. I'm seeing:

```
Undefined variable: skin_author_uri"
in /path/to/textpattern/include/txp_skin.php at line 250.
```

and a load of these on the Themes panel:

```
Notice "Undefined index: title"
in /path/to/textpattern/vendors/Textpattern/Skin/Main.php at line 274.
```

Probably just a collision with my old version since I just overwrote a previous dev install?

NicolasGraph:
The first error could appears because of the DB field change I mad from `website` to `author_uri`. Did you apply this change to your DB in a way or another?

NicolasGraph:
The second one could appears because of the change applied to the theme JSON files: the `name` field was changed to `title`.

Bloke:
I just overwrote everything and logged in. The upgrade script won't of course know that I'd previously used the "wrong" index name so this is probably just an issue with my process rather than the code.

makss:
@Bloke Help me please solve the merge conflict #966
This Pull Request contains the correct files `manifest.json`
```
Conflicting files
textpattern/setup/manifest.json
```
This file needs to be deleted.

Bloke:
I tried to use the online editor to fix the conflict. Did a commit but it's not showing up. Essentially I just removed the 'else' branch since the one in the themes branch is correct as far as I can see.

makss:
@NicolasGraph , I can not install/update a theme if the theme directory has read-only permissions.
```
directory_creation_failure: lock
```

NicolasGraph:
@makss, I will take a look on monday, in the meanwhile try to remove a potential _lock_ directory and  re-install/update. Themes directories are locked on import/export via a _lock_ directory to avoid conflicts; for an unknown reason, it seems yours was not unlock at the end…

NicolasGraph:
Oh, this PR now has conflicts…

NicolasGraph:
In fact @makss, I misundertood your message, sorry.
In fact, it is the expected behaviour, as I said above, themes directories are locked via a _lock_ directory on import/update/export to avoid conflicts. Of course if the _lock_ directory can't be created, the process fail, so having write access the themes directories is required.
If you think that is a problem and think about a better way to proceed, please open a new issue.

makss:
@NicolasGraph I'm sorry I did not look carefully, this is a problem in the branch `themes`.

@Bloke thanks for helping with the merge conflict.

philwareham:
Is this PR now ready for merge, everyone?

NicolasGraph:
> Is this PR now ready for merge, everyone?

It could, however, as it wasn't in the last days, I would just like to ask to @bloke and @bloatware what should be, according to them, the better way to set the new method related parameters. Just to be sure…

I now have the following to create a skin/theme and its relative assets (pages, forms, styles):
```
public function create(
    $name,
    $title = null,
    $version = null,
    $description = null,
    $author = null,
    $author_uri = null,
    $assets = null
) { … }
```
At first I thought some methods would use only some of the skin table fields, but at the end, I introduced some consistency between these methods, so `create()`, `edit()` and `duplicate_as()`all have these parameters.
The only diffference is that `$name` is required for `create()` only and `$assets` is not available for `edit()`.

Do you think I should do it like so:

```
public function create($row, $assets = null) { … }
```
…using the usual `$row` associative array and documenting required fields?
Or something like:
```
public function create($name, $infos = null, $assets = null) { … }
```
…just to reduce the args number?

NicolasGraph:
I finally decided to go with `public function create($row, $assets = null) { … }` which seems pretty obvious with a fresher point of view. It also has the advantage to allow more fields in the future (like `image_uri` or whatever).

philwareham:
OK, just let me pull latest changes form dev into themes so you are working with the latest branch code - that way hopefully there won't be any conflicts.

philwareham:
OK, themes now has all latest amends from dev.

NicolasGraph:
Thanks @philwareham.
I've merged the _themes_ branch into mine; everything should be up to date now.
There are still a few commits to add to make this PR 'mergeable' again, then I would be happy to work on last fixes through dedicated issues/PR.
I'm going to edit the first comment to introduces the way the _Skin_ package should now work.

PR #987, opened 2017-11-07
`Old` -> `legacy`

petecooper:
Minor language tweak in `HISTORY.txt`, more appropriate and fewer negative connotations.


philwareham:
OK, seems fair - there is also another instance on the line below - if you'd like to amend that too as part of PR I'll merge! 😀

philwareham:
Cheers Pete!

Issue #983, opened 2017-11-06
Admin themes: use manifest.json instead of php hardcode


makss:
We already have everything we need.
It is necessary to improve `Textpattern/Admin/Theme` class and clean manifest() function in themes.

bloatware:
Could `announce(_async)` functions be made a little less abstract in the parent class? They are exactly identical in three official themes. Like most parts, actually.

bloatware:
Re `announce`, all it does atm is inject a message with a status in a template. A theme should just provide a template if not happy with the default one. This template could then also be used for client-side announces I had to introduce for async multiple files upload, though there might be a better solution for this.

philwareham:
Done - many thanks yet again @makss. Closing.

PR #988, opened 2017-11-07
Admin themes

makss:
Few sorted the code in the Admin themes. Please test.
Issue #983

It would be nice in function `names()` use constants for the format, but I do not know what to call them better.


phiw13:
This seems to work well, as far as I can tell. That is – everything behaves as it should, both using Hive and my own Sand space theme (after editing the `.php` file). I’m *very* happy to see `announce()` moved out of the theme file and into core, where it belongs.

At first you had me worried a little with the part where you move the  `admin_custom_css` and `admin_custom_js` out of the theme into core. My first worry was that you would force a theme to use the structure / file path (`assets/*`) used by Hive etc. But that appear not to be case - I left that code part in my own theme (in `function html_head()`) and it still works fine.

I’m not sure that part should be done, though. I think the whole code for `admin_custom_css` and `admin_custom_js` should be part of a theme, not core.

Musing - I wonder if it would be possible at one point in the future for the theme CSS file (and the `viewport` meta) to be called **before** the core JS files, as the very first thing in the `<head />`. That would improve performance (JS being a render blocker). Hmm, tricky thing as it probably would entail quite a bit of changes.

makss:
> My first worry was that you would force a theme to use the structure / file path (assets/*) used by Hive etc.

Thanks for the comment, fixed. You can specify path.
```
    function html_head()
    {
...
...
        $this->cssPath = 'myCssFolder';
        $this->jsPath = '';         // Theme root

        return join(n, $out);
    }
```

---

> I left that code part in my own theme (in function html_head()) and it still works fine.

In your theme, you can completely override function `html_head_custom()`.
```
    function html_head_custom()
    {
        // some custom stuff

        return join(n, $out);
    }
```

philwareham:
If you are both happy with this now I'll merge into `dev` and update my theme repos accordingly.

phiw13:
@makss 

> … You can specify path.

Ooh, thank you. I’ll have to play with that when I get I moment.

> … your theme, you can completely override function html_head_custom().

yeah, after commenting I thought something like that would be possible.

--

@philwareham 

A quick test with Hive + a custom stylesheet after Makks last patch, everything still works fine. And a quick test with (hacked) Sandspace also works as expected. As far as I am concerned, this is good to go.

philwareham:
Thanks! Merged now. I'll copy the PHP code over to my theme repos later today too.

bloatware:
It seems to work well (thanks @makss) save for "Admin-side theme" choice is empty now.

PR #994, opened 2017-11-10
Plugin textpack

makss:
Draft for testing.  It seems it works.

I tested on plugins: `com_connect`, `ied_plugin_composer`.

~~**Known bug:** It is necessary to convert from old language codes to new ones. Now it is possible to incorrectly recode.~~


Bloke:
Haven't tested it, but reading the code I love this idea. So to recap:

* When installing / updating a plugin, all strings from the textpack - in all languages - are stored in the plugin table and any packs that match the installed language(s) are installed.
* Thus all languages are always potentially 'available' with suitable strings for the plugin's use.
* When you install / update a language from the Languages panel, it checks the plugin's 'textpack' column and if it contains any strings for that language, they are automatically installed alongside the core strings.

If so, that is, quite simply, genius. It fixes a whole load of annoying problems with textpacks in plugins.

A few observations:

* What happens if the `#@language` identifier is missing? Without this patch, Txp currently treats the unnamed strings as 'default' and installs them into the currently selected _site_ (not _ui_) language. This is arguably wrong, but it does mean that there shouldn't be any untranslated strings. At least for the site language.
* After 763a8113 (and apply any fixes for RTL that we dream up) we shouldn't have any untranslated strings. Does this commit change anything in this patch? Any additional assumptions we can make?
* Is the proper solution to encourage plugin authors to _always_ specify a language identifier? Does that solve anything?

I'm just thinking out loud here:

* `TEXTPATTERN_DEFAULT_LANG` is `en`.
* Txp site uses `fr`.
* The UI of one of the admin users is set to `de`.
* The admin installs a plugin that has `en-gb`, and `ar-dz` and `pt-br` packs in it.
* Three textpacks are stashed in the `txp_plugin` textpack column but none of them match languages that are installed.

What do we want to happen? Leave plugin strings untranslated? Encourage authors to leave one language without an identifier so it is used as default? (but that means we don't know what language it actually is). Or maybe we could introduce a new directive to textpacks: `#@default` as a flag to indicate that if there are no matching languages, it uses the nominated one as default? And if there's no default flag, then it uses `TEXTPATTERN_DEFAULT_LANG` as the default. I kinda like that option, actually.

Bloke:
P.S. For changing the legacy language codes when a Textpack is installed, look at `L10n\Lang->closest()`.

Bloke:
If we're going down the route of using a lookup table, we'd best use it in _to_4.7.0.php as well to update the lang codes in the `txp_lang` table. The `closest()` method might fail in some cases (e.g. Norwegian, Serbian, ...)

makss:
With language packs Textpattern no problems, we will re-encode them once. But with textpacks for plugins it's impossible to do this. Therefore I did `L10n\Locale->validLocale()`

Or another option is not to install a textpack with the wrong language code.

Bloke:
Sure, I was just thinking that your lookup table would be useful during the one-time conversion of languages from the legacy system to the new shorter lang codes, instead of relying on guesswork like is done in the `closest()` method.

PR #1017, opened 2017-11-19
Fix broken code of conduct link

petecooper:


philwareham:
Thanks Pete, missed that before.

PR #1067, opened 2017-12-15
Last Skin package

NicolasGraph:

- grouped queries where possible;
- essential templates management improved (adds missing essential templates on import);
- UI messages handling improved;
- misc. bug fixes.


philwareham:
Thanks Nicolas, appreciate all the work you’ve put in to this. Merged for thorough testing now.

NicolasGraph:
Sorry for the delay and thanks for your encouragements @philwareham, and @Bloke also!
I'll work step by step through dedicated commits from now to be more transparent on the next improvements.

Just to be clear about something which is maybe not obvious, I would like to notice that, on import, if an essential form is not stored in the right type related subfolder, an error is displayed and the form is not imported; an empty one with the right type is created to replace it.

I could just change the type of the form but I would need to be sure that it does not also exists in the right type related subfolder. It is probably something that could be done as a future enhancement. What do you think?

Anyway, feedbacks are welcome.

PR #1092, opened 2018-01-08
2018 year bump

petecooper:
tra la la


philwareham:
Ta! Not sure why my IDE missed those files but cheers for double-checking.

PR #1129, opened 2018-01-24
Multi-site pre-installer and welcome article improvements

jools-r:

- Use `adminurl` as replacement string in welcome article in place of `siteurl/textpattern` (support for `siteurl` is retained although currently unused). Replaces with `//admin.sitename.tld` on multi-site installations and `//www.sitename.tld/textpattern` on regular installations. Tested with both. 
My [adaptation of makss' url replace routine](https://github.com/textpattern/textpattern/pull/1129/files#diff-f7309db96748be1cef1e8a22597532f3R210) in TxpXML.php could probably be made more elegant. I just passed it through twice. **Please review code**.
- Simple ‘Hive’ styling added to the pre-installer. It's not a thing of beauty but not as barebones as before. Rather than extracting the relevant CSS selectors from textpattern.css, I copied it wholesale from the hive admin-theme directory. This makes it easier to update and the file is deleted anyway after installing. We can't, unfortunately, link directly to the actual textpattern.css as we don't know its location at this stage. If you're very keen on keeping the size down, maybe @philwareham could create a second sass build just for the setup screens.

Remember, this is just fallback stuff. In a normal installation with correctly transferred files, you see none of this: it goes straight to the language choice screen.

1 - Textpattern directory not found
txp_multi-site_textpattern-dir-not-found

1b – Textpattern directory: wrong path entered, try again
txp_multi-site_textpattern-dir-wrong-path

2 – Textpattern directory found and symlinks generated. Proceed to regular installation.
txp_multi-site_textpattern-dir-found-symlinks-generated

2b – Textpattern directory found but symlinks not generated. Instructions for linux / unix.
txp_multi-site_textpattern-dir-found-symlinks-not-generated-unix

2c – Textpattern directory found but symlinks not generated. Instructions for windows.
txp_multi-site_textpattern-dir-found-symlinks-not-generated-windows
*(This last window is faked by triggering the error. On a proper windows machine, the forward slashes would be backslashes)*


philwareham:
Sorry, what's the CSS exactly? Is it just a carbon copy of the Hive CSS from my admin theme or a subset of it?

philwareham:
Ah, OK, just read the initial post from @jools-r - we need to build (in my actual Hive repo) and minimal CSS ruleset for this purpose. Please open an issue in the main Textpattern repo and flag it as 4.7 milestone otherwise I may miss this. Thanks.

PR #1153, opened 2018-02-15
Missing commit to update `last_template_saved` on skin switching

NicolasGraph:
Sorry…

### DIFF
--- a/textpattern/vendors/Textpattern/Skin/AssetBase.php
+++ b/textpattern/vendors/Textpattern/Skin/AssetBase.php
@@ -735,7 +735,9 @@ namespace Textpattern\Skin {
             }
 
             if ($skin) {
-                $Skin = $this->getSkin()->setEditing($skin);
+                $Skin = $this->getSkin();
+                $Skin->setEditing($skin);
+                $Skin->setName($skin);
             }
 
             $this->getEditing();


philwareham:
@NicolasGraph @Bloke Do you want to merge `themes` into `dev` for more extensive testing, or do you still want to test in the specific branch? I did a few trials earlier and seems to be working in my limited tests.

Bloke:
Not sure it's quite ready yet. Veeeery close. Haven't tested it under error conditions enough.

Yesterday, it wasn't populating the `txp_section` table with the correct theme that you chose during setup: it was always putting 'default' in the `txp_section` table as the theme name, regardless if I chose `four-point-seven` or one of my own themes during setup. That may be fixed now, but needs verification.

Also, please anyone check that you can still choose and successfully import themes during fresh setup from both the `setup` and your own themes in the `themes` directory.

And @petecooper if you could run a few auto-install tests, that would be awesome, thanks :-)

There are still a few corner cases to figure out. e.g. what happens if you rename Pages and Stylesheets. You can't delete Pages and Styles that are "in use" by other themes, i.e. those assets that have the same name.

Nicolas has helped smooth things in this commit, but there might be the odd semi-unexpected behaviour. Such as if you try to install a theme that does _not_ have `default.txp` in it (Page or Style). They'll be created for you and used in the Section assignment. Same if you remove them and export, they'll be added to the bundle. Sort of, pseudo "essential" Page and Stylesheet. Not sure about that yet but it's tricky to balance. Without some baseline, how do we know which Page in your set of pages is the one to be used for each section during setup? We don't know. We can't possibly know.

The only way we _might_ be able to infer this is if there is _only one_ template in the directory. Then we could perhaps safely assume that the section assignment during setup could be set to use that Page. But that may involve custom codeflows (is this a theme update vs is this installation during setup). Not sure. I think we need to live with this, put it in the beta and see what people think. If it means we have to tweak stuff before RC or final, so be it.

A few other minor things. See 297d3cc7f. Notice I changed the `gTxt()` messages to remove `_duplication` and used `_creation` instead. I think I missed [one](https://github.com/textpattern/textpattern/blob/themes/textpattern/vendors/Textpattern/Skin/Skin.php#L651) or [two](https://github.com/textpattern/textpattern/blob/themes/textpattern/vendors/Textpattern/Skin/Skin.php#L680). If anyone fancies fixing those, that'd be awesome, thanks. IMO, we don't need "duplication of these things you just selected failed/succeeded" messages: we need "creation of these things you attempted to duplicate failed/succeeded". So references to `$name` (thing bneing duplicated) need to be `$copy` (thing being created) instead.

There are also stray variables in use, such as `$notImported` (see the second link I just posted above to line 680) which, as far as I can tell, isn't defined anywhere. Need to make sure there aren't any of these lying around before we merge. That'll require putting sites in debugging mode and making "bad" page names with disallowed characters and stuff to see what happens when you perform mult-edit operations on them.

Finally, I noticed some odd behaviour when selecting multiple themes and performing mulit-edit actions such as 'duplicate'. Select theme A and B, duplicate them. After duplication, Theme A_copy, Theme B_copy and Theme B were selected in the list, but Theme A wasn't. Might have been a glitch: didn't explore further. Anyone confirm or deny this? What is the expected outcome? When duplicating articles, only the original two remain selected. Is that the behaviour we want here too on the Themes panel?

bloatware:
> Finally, I noticed some odd behaviour when selecting multiple themes and performing mulit-edit actions such as 'duplicate'. Select theme B and C, duplicate them. After duplication, Theme B_copy, Theme C_copy and Theme C were selected in the list, but Theme B wasn't. Might have been a glitch: didn't explore further. Anyone confirm or deny this? What is the expected outcome? When duplicating articles, only the original two remain selected. Is that the behaviour we want here too on the Themes panel?

My guess is that `B_copy` has already been selected and deleted, so its selected state was retained and restored after the second `B` duplication. This does not happen with articles/comments/files/images/links that are indexed by their unique ID. That's also why I hesitated to extend async lists beyond Content tabs.

Bloke:
@bloatware You may well be onto something there. We don't necessarily need async here - the number of themes is likely to remain pretty small in most people's projects so the overhead of a page refresh isn't onerous. Nicolas just threw it in.

We can always revert to regular page wrangling for now. When we (ahem, you!) rejig `textpattern.js` in future versions, we may be able to roll this out across the board.

What do you think @NicolasGraph ? Is it easy enough to switch back to normal table list mode on the Themes panel, or is it a lot of work?

bloatware:
I can just add some parameter to `Table->render()` function to disable `async` mode, so the code change would be minimal.

NicolasGraph:
While it would be possible to go back, I like the @bloatware suggestion! ;)

Bloke:
Like! Like!

bloatware:
[Seems to work](https://github.com/textpattern/textpattern/commit/9656d332e20f7b698ab42e45f40706b07cc0a199)

Issue #1160, opened 2018-02-20
Automated installer front-side theme - `default`, with no content


petecooper:
(Current 4.7.0-dev)

### Expected behaviour

Automated setup with config for `Four Point Seven` theme should install and use `Four Point Seven` theme.

### Actual behaviour

Automated setup with config for `Four Point Seven` theme installs a theme called `default` with no content, despite console output saying `Four Point Seven` was installed.

Possible regression since themes branch was merged in?

### Steps to reproduce

Using this JSON:

```
{
    "site": {
        "lang": "en",
        "siteurl": "",
        "theme": "",
        "public_theme": "/setup/themes/four-point-seven",
        "datadir": ""
    },
    "mysql": {
        "user": "dev-demo",
        "pass": "",
        "host": "127.0.0.1",
        "db": "dev-demo",
        "table_prefix": ""
    },
    "user": {
        "realname": "Textpattern Demo Robot",
        "email": "domo-arigato-mr-roboto@example.com",
        "name": "textpattern-robot",
        "pass": ""
    }
}
```

…gives this console output:

```
Checking connectivity... done.
[OK]	Connected
[OK]	Using dev-demo. (utf8mb4)
[OK]	Creating database tables
[OK]	Lang: 'en'
[OK]	Prefs: 'data/core'
[OK]	Import: 'data/txp_category'
[OK]	Import: 'data/txp_link'
[OK]	Import: 'data/txp_section'
[OK]	Import: 'articles/articles.welcome'
[OK]	Front-side theme: '/setup/themes/four-point-seven'
[OK]	That went well!
```

…and this in the admin-side:

![safariscreensnapz001](https://user-images.githubusercontent.com/226578/36442034-5b1e3666-166c-11e8-9381-97b079d5db2c.png)

Tagging https://github.com/textpattern/textpattern/issues/1158 and https://github.com/textpattern/textpattern/issues/1101 for completeness.

Bloke:
@NicolasGraph - eeek! Any ideas?

phiw13:
Hmm, I’m not sure this is a regression from the most recent _themes_ refactoring. I noticed the same thing recently on a 2~2½ week old install (manual, using _setup_).

I had installed that for testing something completely unrelated, did my tests but never checked anything under Presentation at the time. I only noticed it more than a few days later, and wasn’t sure anymore what actions  I had taken previously (and then I forgot about the whole thing… flue and high fever terrorizing my body…).

[...]

Issue #1158, opened 2018-02-20
Automated installer defaults to `zero` theme


petecooper:
I'm rebuilding the demo server from scratch in anticipation of the upcoming beta, and it looks like the automated installer is defaulting to the `zero` theme when no theme is explicitly requested in the JSON file.

Tagging in https://github.com/textpattern/textpattern/issues/1101 for completeness.

**JSON with no theme**

```
{
    "site": {
        "lang": "en",
        "siteurl": "",
        "theme": "",
        "public_theme": "",
        "datadir": ""
    },
    "mysql": {
        "user": "dev-demo",
        "pass": "<snipped>",
        "host": "127.0.0.1",
        "db": "dev-demo",
        "table_prefix": ""
    },
    "user": {
        "realname": "Textpattern Demo Robot",
        "email": "domo-arigato-mr-roboto@example.com",
        "name": "textpattern-robot",
        "pass": "<smipped>"
    }
}
```

This connects fine, and recognises English, and then installs `zero` (ignore the `beta.sh` name, I'm still working on it):

```
$ sudo bash ./beta.sh 
Cloning into '/var/www/vhosts/textpattern.co/dev-demo/staging/github-dev'...
remote: Counting objects: 467, done.
remote: Compressing objects: 100% (419/419), done.
remote: Total 467 (delta 99), reused 120 (delta 21), pack-reused 0
Receiving objects: 100% (467/467), 1.35 MiB | 0 bytes/s, done.
Resolving deltas: 100% (99/99), done.
Checking connectivity... done.
Cloning into '/var/www/vhosts/textpattern.co/eu/dev-demo/staging/github-dev'...
remote: Counting objects: 467, done.
remote: Compressing objects: 100% (419/419), done.
remote: Total 467 (delta 99), reused 120 (delta 21), pack-reused 0
Receiving objects: 100% (467/467), 1.35 MiB | 0 bytes/s, done.
Resolving deltas: 100% (99/99), done.
Checking connectivity... done.
[OK]	Connected
[OK]	Using dev-demo. (utf8mb4)
[OK]	Creating database tables
[OK]	Lang: 'en'
[OK]	Front-side theme: 'zero'
[OK]	That went well!
```

**JSON with theme title**

Note: I've changed the language to `fr` as a visual cue to the installer reading the correct JSON file.

```
{
    "site": {
        "lang": "fr",
        "siteurl": "",
        "theme": "",
        "public_theme": "Four Point Seven",
        "datadir": ""
[...]

PR #1161, opened 2018-02-22
Skin enhancements

NicolasGraph:
- Moves `isDirEmpty()` to txplib_misc.php as `is_dir_empty()`;
- Do not accept unknown public theme name: related to https://github.com/textpattern/textpattern/issues/1160 + https://github.com/textpattern/textpattern/issues/1158.

### DIFF
--- a/textpattern/lib/txplib_misc.php
+++ b/textpattern/lib/txplib_misc.php
@@ -7036,3 +7036,24 @@ function JSONPrettyPrint($json)
 
     return $result;
 }
+
+/**
+ * Wether a directory is empty or not.
+ *
+ * @param  string $path The directory path
+ * @return mixed        NULL if the directory is not readable (or does not exist),
+ *                      TRUE if empty, otherwise, FALSE.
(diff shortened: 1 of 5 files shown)


NicolasGraph:
As this method job was pretty "generic" I just thought that it could be placed outside the _Skin_ package so other devs could easily see it but if you prefer to keep protected it in the package for now, just let me know. It is maybe not time to feed old school files?

Bloke:
Good thinking about the generic nature of the method, but I agree that we should try to limit what's added to `txplib_misc.php` and start to shrink it in favour of using proper classes in future. I'm guilty of adding stuff to it in recent memory because it's easy, but I'm trying to break myself out of the old school mindset!

NicolasGraph:
Ok, I'll bring it back into the package. I also put `JSONPrettyPrint($json)` in _txplib_misc.php_ few weeks or months ago and it does not seem to be used in other places, I could also protect it into the package for now.

PR #1168, opened 2018-02-27
Avoids empty theme titles in import forms (install/skin tab) 

NicolasGraph:
Use the theme name with first character capitalized on empty title.


textpattern/vendors/Textpattern/Skin/Skin.php
```diff
@@ -239,22 +239,31 @@ protected function getFileContents()
...
-            if ($contents !== null) {
-                extract($contents);
+            extract($contents);
 
-                !empty($title) or $title = ucfirst($this->getName());
-                !empty($version) or $version = gTxt('unknown');
-                !empty($description) or $description = '';
-                !empty($author) or $author = gTxt('unknown');
-                !empty($author_uri) or $author_uri = '';
+            return $this->parseInfos($title, $version, $description, $author, $author_uri);
```
Bloke: @NicolasGraph Sorry, I don't get it. Surely if the `json_decode()` fails for whatever reason (bad path, mangled content, ...) then the `extract()` won't extract all the variables properly, and then this line will throw a bunch of warnings about uninitialised variables? Or have I missed something?

NicolasGraph: You're right, sorry, I'll fix it.

Issue #908, opened 2017-06-29
Introduces plugin help files rendering


NicolasGraph:
As discussed [here](https://forum.textpattern.io/viewtopic.php?id=47833), the fact that the Textpattern uses Textile in restricted mode to parse plugin help files limit ways to style them. I used to apply Textile on compilation to be free to customize my files by adding tabs to avoid a too long page of text but it does not work when plugins are installed via Composer.

Following a @Bloke [recommendation](https://forum.textpattern.io/viewtopic.php?pid=306055#p306055), I worked on a way to style plugin help files on reading. It would have the advantage to simplify the help files creation and would allow to keep a nice design consistency in the future. 

For now, to keep supporting not well strutured files, the script check if the first help element is a `h1` tag and if there are 2 or more `h2` tags — we don’t need one tab… — before to change the layout. Otherwise, everything stays as is. Then, any contents between the `h1` and the first `h2` tag is displayed just under the header; it is usually a plugin description. After that, each `h2` is read as the beginning of a new section which is linked to a tab as for pref groups, and in the end, the Prism related `.language-markup` class is applied on code chunks

You can already try it with plugins such as [ied_plugin_composer](https://github.com/Bloke/ied_plugin_composer) or [smd_article_stats](https://stefdawson.com/sw/plugins/smd_article_stats) — and probably other Stef's plugins (?).

![Plugin help files rendering demo](http://g.recordit.co/qAnkcyC0o5.gif)

I'd be glad to send a pull request — should I? — if you are interested in but before, we should at least rename `.txp-prefs-group` to something more generic. Any idea @philwareham? I already changed some variable names.

Oh… yes, I also had to change the way the local storage works on the plugin tab by using plugin names as data items to keep tracks of each last visited tabs.

---

### Tested plugins

Here are some test results (checked when ok, default rendering otherwise):
- [x] _abl_droploader_
- [x] _adi_article_tab_
- [x] _adi_list_
- [x] _arc_meta_
- [x] _arc_vimeo_ (an existing table of contents would need to be removed as it display a useless tab)
- [ ] _cbe_frontauth_ (contains some styles)
- [x] _com_connect_
- [ ] _cwh_link_to_home_ (does not contain any secondary header/section)
- [ ] _cxc_templates_ (contains some javascript)
- [x] _etc_pagination_
- [ ] _etc_query_ (does not contains any `<h1>` tag + other HTML structure issues)
- [x] _ied_plugin_composer_
[...]

PR #1171, opened 2018-02-28
Plugin help rendering

NicolasGraph:
## Description

These changes improve the default plugin help rendering by using some tabs as for prefs (See https://github.com/textpattern/textpattern/issues/908.).
The plugin help contents structure must be the following to benefit of this feature:

```
<h1>Unique first-level header</h1>

Any kind of optional presentation.

<h2>First tab/section</h2>

…

<h2>Second tab/section</h2>

 … 
```

It also need to be free of `<script>` and `<style>` tags and should not use any `style` attribute for now. Classes are OK as long as they are not related to the layout (`.txp-layout*`, `.txp-grid*`).  

## Optional improvements

@philwareham, you may want to change `.txp-prefs-group` [here](https://github.com/textpattern/textpattern/compare/dev...NicolasGraph:plugin-help-rendering?expand=1#diff-022350421a5c4d37046c7e91782e5438R2274) and [there](https://github.com/textpattern/textpattern/compare/dev...NicolasGraph:plugin-help-rendering?expand=1#diff-022350421a5c4d37046c7e91782e5438R2291) to something more generic such as `.txp-tab-contents`, or add a dedicated class such as `.txp-plugin-help-section`.

Also, I use two textpack related strings [here](https://github.com/textpattern/textpattern/compare/dev...NicolasGraph:plugin-help-rendering?expand=1#diff-022350421a5c4d37046c7e91782e5438R2276) and [there](https://github.com/textpattern/textpattern/compare/dev...NicolasGraph:plugin-help-rendering?expand=1#diff-022350421a5c4d37046c7e91782e5438R2297). While these existing strings are OK, they could be replace by dedicated ones.

It is still allowed to use the `id` attribute as it can be useful for anchor links. Should we forbid `id` values already in use such as `page-plugin`, or be confident with plugin author to work properly?


NicolasGraph:
The previous commit disable the rendering if the help file contains any class (except `.caps` that is use for abreviations). We could also allow the use of few more useful classes such as `.alert-block.information`.

philwareham:
@bloke OK to merge?

Issue #1194, opened 2018-03-15
Incorrect HTML structure / markup on the Themes panel


phiw13:
The multi-edit widget (`div.multi-edit`) is part of the `div.txp-navigation` (inside). It should precede it as on any other list panel, and inside the `.multi_edit_form` form. The `div.txp-navigation` should follow the `form`.

Themes panel:

```
form.multi_edit_form
  > div.txp-navigation
    > div.multi-edit
```

Sections panel:

```
form.multi_edit_form
  > div.multi-edit
/form
div.txp-navigation
```

2 screenshots:
https://dev.l-c-n.com/_junk/Sections.png
https://dev.l-c-n.com/_junk/Themes.png

NicolasGraph:
Thanks for the report @phiw13.

philwareham:
This should fee fixed now - let us know. Cheers all.

phiw13:
Yup, looks much better. Thanks @NicolasGraph .

PR #1195, opened 2018-03-15
Fixes incorrect HTML structure / markup on the Themes panel

NicolasGraph:
See https://github.com/textpattern/textpattern/issues/1194.


philwareham:
Thanks!

Issue #1199, opened 2018-03-20
Theme 'override' - rethink wordage


philwareham:
The following option 'Override' in the new themes panel is too obtuse and confusing to end users (me included):

![grab1](https://user-images.githubusercontent.com/413665/37647592-8580e872-2c24-11e8-88aa-4498ed17cf3a.png)

It would be clearer if the wording was as follows I think:

![grab2](https://user-images.githubusercontent.com/413665/37647613-93f145a0-2c24-11e8-822e-831c3c32091f.png)

Any objections @bloke and @NicolasGraph ?

Bloke:
Fine by me.

ziselive:
Why doesn't my theme management page import theme options?

NicolasGraph:
@ziselive, I guess you don't have any new uploaded theme in your themes directory.

NicolasGraph:
It's ok for me too @philwareham.

Bloke:
> Why doesn't my theme management page import theme options?

@ziselive What theme options are you referring to? Prefs? Information stored in the manifest? Themes don't natively import prefs right now.

@NicolasGraph We could leave the string key as `override` for now, coudn't we? And just change the string representations in languages (plus pophelps)?

philwareham:
@Bloke it's same amount of upheaval even if we keep `override` string. Personally I'd say keep `override` and move it to the `common` group and also add `synchronize` in `common` too. They seem fairly generic and useful words to have available.

bloatware:
Even in English `resynchronize` sounds weird for me (the spell checker confirms), and in Russian it would be abracadabresque. It is also bidirectional: synchronize in which sense? Couldn't we find something better to say "reload from disk"?

philwareham:
@bloatware I believe the sync **is** bidirectional - if unused items are in database and not flat files, remove those / if unused items are in flat files and not database, remove those.

ziselive:
@NicolasGraph you're right.  @Bloke ，I don't have any new uploaded theme in  themes directory. Put in a new theme and display it.

Bloke:
If you choose this option (what essentially amounts to a "reimport" or "update" of what's in the DB) then the deletion is one way. In other words:

* If you're importing from disk and the checkbox is on, the _database_ templates are deleted and loaded from disk.
* If you're exporting from the DB and the checkbox is on, the _disk_ templates are deleted and recreated from the database contents.

NicolasGraph:
[...]

PR #1200, opened 2018-03-20
 'Override' option to 'Resynchronize'

NicolasGraph:
See https://github.com/textpattern/textpattern/issues/1199.


philwareham:
Thanks. Needs lowercase `r` on the string name though, like so: `gTxt('resynchronize')`.

PR #1225, opened 2018-04-03
Proof read `HISTORY.txt`

petecooper:

- non-markup theme elements (images etc) are not stored in the database
- Cmd+S works for article Publish and Save, also
- clarified form recursion
- reworded a couple of `now accepts` instances for uniformity – it's clear from the bullet header there's a change, so `now` is unnecessary
- caps case on `Pages`, `Styles` and `Forms`


philwareham:
Thank you once again!

Issue #1264, opened 2018-05-10
Should logging include homepage/root URLs?


petecooper:
### Expected behaviour

Homepage hits appear in Logging as a `/` or similar.

### Actual behaviour

Homepage hits appear in Logging as an empty cell.

### Steps to reproduce

* Enable all Logging
* Visit various pages and URLs within the site
* Note nothing is recorded for the URL in Page column in Visitor Logs

#### Additional information

Empty cells in the following screenshot represent homepage/root/`default` hits:

![safariscreensnapz001](https://user-images.githubusercontent.com/226578/39871991-7eae852e-545e-11e8-8319-931896a597ad.png)

petecooper:
(Follow up – I don't use internal logging at all, so this might be how it's always been, please instaclose if that's the case.)

philwareham:
Yeah, I noticed that too a couple of days back (don't use logging myself - just turned it on for GDPR testing). I should I think show something, looks weird without it - just don't know what. Not sure if forward slash would work (probably would have to append it to the child URLs too)?

Bloke:
Hmm, yeah, that is odd. We could use the URL for the "home page" but that might then look odd alongside all the other pages that don't have it. And putting it on all pages just clutters the panel.

Maybe just use something we have a `[common]` gTxt string for, and put it in square brackets or summat. Like:

[Home]

or

[Site]

??

philwareham:
Changing line 262 of `txp_log.php` from:

    $log_page = href(txpspecialchars($log_anchor), txpspecialchars($log_page), ' target="_blank"');

To:

    $log_page = href('/'.txpspecialchars($log_anchor), txpspecialchars($log_page), ' target="_blank"');

Adds a preceding `/` to all entires which *may* fix this. What I don't know is if URLs are relative to the domain root or to the root of the Textpattern location?

Bloke:
That works and I can't see a downside: it's only the visible part of the link that's being modified so shouldn't affect the destination in any way. We'll have to get someone to test this in a multi-site setup to see what happens there. @jools-r ?

jools-r:
Almost works. The visible part of the link acquires a preceding slash, so the home page is now `/`, but the href of the link itself is not relative to the front side link. It needs to be:

```
$log_page = href('/'.txpspecialchars($log_anchor), rtrim(hu,'/').txpspecialchars($log_page), ' target="_blank"');
```
[...]

PR #1273, opened 2018-05-13
Homepage in visitor logs + correct front-side links

jools-r:
As per #1264 raised by @petecooper and proposed by @philwareham.

- Forward slash prepended to display of pages visited in the visitor logs
- Homepage appears as '/'
- Correct front-side link in regular and multisite installations


philwareham:
Thanks (Stef used your code)

PR #1283, opened 2018-05-22
Skin update related to the PHP min. requirement change

NicolasGraph:
Hi, Comming back from holidays I see minimum requirements have changed; here is a Skin package related update to use the `JSON_PRETTY_PRINT` flag on `json_encode()` instead of a custom method. It could also be used in `TEXTPATTERN_JSON` directly but I don't think that it is useful for now.


NicolasGraph:
Ok @bloatware.

Issue #1290, opened 2018-05-29
Add/include Nginx directives to (optionally) prevent access to assets


petecooper:
(Please assign this to me for now and throw a 4.7.2 milestone on it for good measure. Thank you.)

Need Nginx `locations` directives for:

- [ ] https://github.com/textpattern/textpattern/blob/dev/themes/.htaccess
- [ ] https://github.com/textpattern/textpattern/blob/dev/files/.htaccess-dist

PR #1292, opened 2018-05-29
Create nginx-dist.conf

petecooper:

- First pass of an Nginx scaffold. Includes directives taken from https://github.com/textpattern/textpattern/blob/dev/themes/.htaccess and https://github.com/textpattern/textpattern/blob/dev/files/.htaccess-dist

Subject to further testing, this resolves https://github.com/textpattern/textpattern/issues/1290

### DIFF
--- a/nginx-dist.conf
+++ b/nginx-dist.conf
@@ -1,0 +1,36 @@
+upstream php-fpm {
+  server 127.0.0.1:9000;
+}
+
+server {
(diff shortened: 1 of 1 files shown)


petecooper:
Actually, thinking about this further – is this really needed as a file? It can be spun out to the docs site easily enough, and carries more space for explanation. I'll close.

PR #1307, opened 2018-06-22
Update README.md

petecooper:

- More sys req table tidying and clarification.

Wondering whether this might be better house in docs repo instead…


philwareham:
Thanks again!

PR #1308, opened 2018-06-26
Update README.md

petecooper:

- Tweak Nginx requirements for 4.8

Textpattern requirements on Nginx are pretty light, overall. I've successfully tested back to to 1.10 (2016) and all is well, so I would be inclined to set that as the lower benchmark and see what feedback we get.

There are two Nginx release streams: mainline and stable. Mainline is preferable to stable as it receives all bug fixes and security updates.


philwareham:
Nice!

PR #1333, opened 2018-09-05
Fixes $this in static Skin related methods

NicolasGraph:

- keep `setDir()` static but do not return `$this`;
- make `resetEditing()` an object related method to support `$this`.

Note that both methods are protected and not in use for now so they are not that useful…


philwareham:
@Bloke is this PR still valid? Thanks.

Bloke:
I think so. The only reason I didn't merge it before was because it had conflicts. Looks like Nicolas resolved them so it's probably okay to merge.

NicolasGraph:
> Looks like Nicolas resolved them so it's probably okay to merge.

Yes, conflicts were fixed.

Bloke:
Thank you. Merged.

PR #1378, opened 2018-12-20
Update txp_diag.php

petecooper:
Add diags check for GD WebP support.

Textpattern image support untouched, this is a preflight check.


petecooper:
screen shot 2018-12-20 at 17 11 51

Issue #1377, opened 2018-12-19
`Show password` on change password area only shows new password


petecooper:
### Expected behaviour

Checking `Show password` on the password change area should show me my existing and new password, so I can check I have spelled both correctly.

### Actual behaviour

Only new password is shown in clear text.

### Steps to reproduce

* log in admin side, change password, observe result of checking `Show password`.

<img width="874" alt="screen shot 2018-12-19 at 16 20 57" src="https://user-images.githubusercontent.com/226578/50233051-755f1500-03aa-11e9-9117-efe30cc27cca.png">

#### Additional information

Textpattern version: current `dev` branch

PR #1380, opened 2018-12-29
Add `txp-maskable` input class to current password

petecooper:
See https://github.com/textpattern/textpattern/issues/1377 for background.

Adds a `txp-maskable` input class to current password field. Nothing else.

This should resolve the above issue, though there's a semantics question - should `Show password` be plural since there are two passwords on the panel?

Alternative route might be to have an additional `Show password` checkbox to toggle each input individually - any strong feelings on that @philwareham ?

### DIFF
--- a/textpattern/include/txp_auth.php
+++ b/textpattern/include/txp_auth.php
@@ -164,7 +164,7 @@ function doLoginForm($message)
                     array(
                         'name'         => 'p_password',
                         'autocomplete' => 'current-password',
-                    ), '', '', '', '', INPUT_REGULAR, '', 'login_password', false, true),
+                    ), '', 'txp-maskable', '', '', INPUT_REGULAR, '', 'login_password', false, true),
                 'password', '', array('class' => 'txp-form-field login-password')
             ).
             graf(


Bloke:
I don't have an overarching reason, but using the checkbox to reveal both passwords feels... Odd. Everywhere we use this checkbox it only affects the associated input. Here, it's different, which is perhaps unexpected for users.

Pluralising the sentence seems a waste of effort for little gain. Adding a dedicated checkbox for the old password feels like overkill. I dunno. Revealing the old password seems redundant to me: you know it or you don't, and if you don't then you have no business being on the page to change it.

But I get that it could be a time saver if you have a really complicated pass and need to check it, so let's leave this as it is now, in the absence of anything better.

petecooper:
>Revealing the old password seems redundant to me […]

Morning @Bloke –

Pure a typo check in my mind. Input is confirmed during or after entry, not revealed from a known-correct password picked from the database. Totally agree on the pluralising and extra check box.

philwareham:
Happy New Year guys! Maybe I accepted the pull request a bit hastily. Feel free to revert if you feel it could be handled differently.

Issue #1142, opened 2018-02-06
Inline JS and CSS and Content Security Policy


philwareham:
Currently, the reliance on inline JavaScript and inline CSS makes it near impossible to create a secure [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) for Textpattern.

Sure, you can set strict rules in your root `.htaccess` file that disallow inline JS and CSS for the site itself but you then need to create another `.htaccess` in your `textpattern` subdirectory with `'unsafe-inline'` directives set to get Textpattern's control panel to work. That takes away a large chunk of the security provided by CSP.

If possible we need to move core JavaScript to external files and use classnames instead of inline CSS `style` attributes.

Obviously plugin authors are a separate issue, but at least we can then say the core can be covered by CSP.

Bloke:
Inline JS is something we should be 90% of the way to designing out. Oleg's done some great work in `textpattern.js` this release to let us use the Routing capability in a more flexible manner. Plus there's the whole "rendering stuff in a View" (`Admin/Table.php` for starters) that'll be coming online, so panel generation and visualization will be more centralized. That should cut down a lot of variation and allow us to leverage reuse more.

There was a bit of inline JS to handle the search widget, but that may have gone now, not sure. There's a line or two of inline CSS in `txp_diag.php` that we should ideally get rid of now. Can't really do much about the public tags that take a `style` attribute and therefore, by inference, the tag builder. Beyond that I haven't paid much attention to the specifics. Are there places in particular that you see inline CSS/JS that we need to stamp out?

philwareham:
In the `<head>` the inline `<script>` that defines `var textpattern` is a biggie. As is the `<script>` in the `<footer>`.

CSS is mainly stuff like `style="display:none"` which could and should be replaced with a suitable `class` and so on. CSS isn't really as crucial as JS as it's less of a security risk but if we can shore it up we should.

bloatware:
The main source of inline scripts are imo the dynamic parameters/preferences/strings that we need to pass to static `textpattern.js`. Dunno what is the best solution here. We certainly could serve JS files via PHP, but it would mean extra processing time.

More globally, `textpattern.js` is becoming a mess (mea culpa), so something must be done in 4.8 anyway.

philwareham:
I'll mop up some of the easy CSS ones.

Bloke:
[...]

PR #1406, opened 2019-05-06
Added some security HTTP headers

AmritasyaPutra:
These values can be applied by default to all textpattern installation. Unless there is a specific plugin that explicitly needs these security settings to be relaxed/tweaked these can help improve the security of a default installation. 'unsafe-inline' in csp of script-src is a reminder that it can be eventually fixed to improve security. I use these and more tweaked settings with reporting on my website. The CSP settings can be verified with csp-evaluator.withgoogle.com.

- Add HTTP security header by default.
- Add CSP HTTP header.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -24,4 +24,15 @@
     AddType image/svg+xml  svg svgz
     AddEncoding gzip       svgz
 </IfModule>
+
+<IfModule mod_headers.c>
+    Header always set X-Frame-Options DENY
+    Header always set X-Content-Type-Options nosniff
+    Header always set Referrer-Policy strict-origin-when-cross-origin
+    Header always set X-XSS-Protection "1; mode=block"
+    Header always set X-Permitted-Cross-Domain-Policies "none"
+    Header always set Feature-Policy "camera 'none'; geolocation 'none'; microphone 'none'; payment 'none'"
+    Header always set Content-Security-Policy "default-src 'self'; frame-ancestors 'none'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline'; object-src 'none'" 
+ </IfModule>  
+  
 # END Textpattern


philwareham:
Thanks for getting involved!

Related: https://github.com/textpattern/textpattern/issues/1142

While I agree people should be using security headers I don’t think it’s within the scope of Textpattern to actually enforce a set of rules. It potentially creates a lot of support issues when users sites don’t work because we’ve blocked a resource they use.

So if we were to include this code I would like to see it commented out with some basic instructions about using at own risk.

petecooper:
Hi @AmritasyaPutra - I say this as a someone who's spent many hours devising and tuning CSP directives across numerous sites – and they're all different – a one-size-fits-all sadly isn't possible. @philwareham is absolutely correct regarding support load – a lot of people won't understand how `.htaccess` works, will see things on their site not working (e.g. embedded YouTube video, analytics, etc) and blame Textpattern.

There's a work-in-progress document outline (here: https://github.com/textpattern/textpattern.github.io/blob/master/installation/content-security-policy.md) with some CSP-related content here, I would personally prefer to see more advisory notices and examples in our docs than `.htaccess` entries (commented or otherwise). Plus, those of us who use Textpattern outside of Apache will be able to adapt the rules where appropriate.

We also have an open issue to improve the CSP foundations here: https://github.com/textpattern/textpattern/issues/1142

AmritasyaPutra:
I think you are right Pete. I implemented a super tight CSP for six of my sites and yes they will cause some features to fail silently if enabled at a later date. This enhancement request can be closed. I will help with the documentation.

philwareham:
Thanks both! Any updates the documentation are appreciated.

bloatware:
Is there any reason why `Options -Indexes` is not enabled by default, though it's more about privacy than security?

philwareham:
@bloatware in my opinion that rule shouldn't even be in our `.htaccess` file *at all*, or any others that are not essential to making Textpattern work. Rather we should add comments that guide users to a more thorough (and documented) resource for further config rules such as https://github.com/h5bp/server-configs-apache/blob/master/dist/.htaccess or a further document in our user documentation site.

bloatware:
@philwareham we do already set (via PHP) some CSP headers, even if they are not mandatory. And we do include `Options -Indexes` in `.htaccess-dist` for `files`, which looks reasonable. It's fine to include a reference, but I don't see why basic privacy/security rules wouldn't be enabled out of the box.

Bloke:
I thought `Options -Indexes` prevented someone from listing a directory contents in the browser? Pretty vital in subdirs like `/images` and `/files` and `/themes` but, arguably, not of much use in the root `.htaccess` file since `index.php` always takes precedence and is executed, isn't it?

bloatware:
@Bloke not sure, if you try to access [https://textpattern.com/files/](https://textpattern.com/files/), you seem to be blocked by Apache, not PHP. And that's another point that bothers me: eventual `files`, `images` and `themes` sections are incompatible with clean modes in the default setup.

bloatware:
And now that we are going to store plugins in fs, there will be another folder to protect.

Bloke:
Ah, I'm getting confused then. I thought we were discussing the `.htaccess` Apache file which should block directory listings out of the box, imo. I'll have to read the thread more carefully sorry.

We've always (well, mostly always, since about 4.01?) had the problem that sections cannot be named the same as one of the files/dirs in root. There's even a warning in Diagnostics about it if you do it by mistake.  Just one of those things, I think. Not sure we can work around it without finding some way for Apache/Nginx to process `index.php` in preference to a physical directory.

bloatware:
No, you are right, I'm just thinking of several things at once. Yes, root is ok, but its subdirectories, say, `textpattern/tmp` are listable.

petecooper:
>I thought `Options -Indexes` prevented someone from listing a directory contents in the browser? 

It does.

>if you try to access https://textpattern.com/files/, you seem to be blocked by nginx, not PHP.

Correct. Nginx doesn't use `.htaccess`.

Bloke:
Cool. I'm not sure about leaving `/tmp` listable out of the box. I suppose there could be something sensitive that lands in there.

It should be moved anyway for people who care about security, but we could probably add directives to its own `.htaccess` file and Nginx config to have it unlistable by default if we deemed it appropriate.

bloatware:
What stops us from enabling `Options -Indexes` in root? It would then be applied to subdirectories too.

Bloke:
*shrug* I would have thought someone would have done that already if it worked and didn't kill anything?! Might explain why it's commented out: perhaps it broke something and someone left it there as a reminder to investigate one day.

It shouldn't interfere with `/textpattern` and `/textpattern/setup` as they have their own `index.php` files so I can't see a problem with doing that off the top of my head. Fewer config files in subdirs too.

philwareham:
My thinking is surely modern servers already implement `Options -Indexes` in the `httpd.conf` file (its ideal location) and if they don't then we should link to resource that not just explains not just this security measure but many other security rules. IMO we should just supply the `.htaccess` rules that make Textpattern work.

Bloke:
@philwareham Makes sense I guess. If we can get away with not specifying it at all and relying on the server config, great. Not really looked into what's default in `httpd.conf` these days as it mostly just works out of the box for my needs. #Lazy.

philwareham:
`Options` default was changed from `All` to `FollowSymLinks` starting with Apache 2.3.11 which was around 8 years ago - so I think it's fairly safe to assume this is nothing to worry about.

bloatware:
FWIW, I have tested few forum users servers (sorry), most of them forbid listing, but not all, even on Apache 2.4. We should at least protect `tmp`, but it is configurable...

> We've always (well, mostly always, since about 4.01?) had the problem that sections cannot be named the same as one of the files/dirs in root.

May be it's time to change their default locations, moving them to some root subdirectory?

petecooper:
I will respectfully suggest (heck, politely recommend, even) we spin this discussion out to individual GitHub issues as a lot of useful ground is being covered and we're creeping away from the PR.

philwareham:
I've cleaned up the .htaccess files now - hopefully that suffices. Feel free to amend if you see fit.

PR #1411, opened 2019-07-01
Use smtp_from preference in txpMail() for sender

AmritasyaPutra:
I expected "SMTP envelope sender address" configuration value to be used to send the email. When it is used for smtp.mailfrom, Return-Path and envelope-from then why not for From field.

If "SMTP envelope sender address" is configured then it should be used as sender. Otherwise existing logic of using the logged in user or destination email address as the sender email address will remain. In today's scenario this means all these emails will be treated as spam as they will fail validation since the textpattern server may not have the authorization to send email on behalf of other domains.

### DIFF
--- a/textpattern/lib/txplib_misc.php
+++ b/textpattern/lib/txplib_misc.php
@@ -2776,25 +2776,34 @@ function txpMail($to_address, $subject, $body, $reply_to = null)
 {
     global $txp_user;
 
-    // Send the email as the currently logged in user.
-    if ($txp_user) {
-        $sender = safe_row(
-            "RealName, email",
-            'txp_users',
-            "name = '".doSlash($txp_user)."'"
-        );
-
-        if ($sender && is_valid_email(get_pref('publisher_email'))) {
-            $sender['email'] = get_pref('publisher_email');
-        }
+    $smtp_from = get_pref('smtp_from'); 
+	$site_name = get_pref('sitename');
+    
+    if($smtpFrom) {
+	$sender = array ("RealName" => $site_name,
+						"email" => $smtpFrom);
     }
-    // If not logged in, the receiver is the sender.
     else {
-        $sender = safe_row(
-            "RealName, email",
-            'txp_users',
-            "email = '".doSlash($to_address)."'"
-        );
+        // Send the email as the currently logged in user.
+        if ($txp_user) {
+            $sender = safe_row(
+                "RealName, email",
+                'txp_users',
+                "name = '".doSlash($txp_user)."'"
+            );
+
+            if ($sender && is_valid_email(get_pref('publisher_email'))) {
+                $sender['email'] = get_pref('publisher_email');
+            }
+        }
+        // If not logged in, the receiver is the sender.
+        else {
+            $sender = safe_row(
+                "RealName, email",
+                'txp_users',
+                "email = '".doSlash($to_address)."'"
+            );
+        }
     }
 
     if ($sender) {


philwareham:
@bloatware @Bloke can you check this one?

AmritasyaPutra:
I had to make correction in my first commit. I think I made mistakes when typing on github, I could have pasted directly from my working setup earlier. Consider these three commit as one, that was the intent. The code works on my setup https://ekvastra.in

bloatware:
I'd leave it with @Bloke since I have no idea why email is working the way it does.

Bloke:
Haha, I was gonna say leave it with @bloatware since I have no idea how email works the way it does :) This patch sounds reasonable though. I'm not sure why it's not using that setting now, I'd have to check the logic of what it does and how it uses it.

We do need to revisit the whole email thing and support SMTP gateways. I wonder if this is the catalyst to jumpstart work on that.

AmritasyaPutra:
I would love to be of help.  I have spent few months implementing SPF, DMARC, SMTP-TLSRPT, MTA-STS correctly on my server, testing with several tools that they work properly and I have been monitoring the reports regularly. Unless these are implemented almost invariable the mails will be marked as spam in today's world.

vanmelick:
On 01/07/2019 20:21, Jyoti Sharma wrote:
>
> I expected "SMTP envelope sender address" configuration value to be 
> used to send the email. When it is used for smtp.mailfrom, Return-Path 
> and envelope-from then why not for From field.
>

Because the email address in the From header is NOT by definition the 
same as the envelope sender address. Back when I added this setting it 
was perfectly valid to have them be different... and it still is today.

> If "SMTP envelope sender address" is configured then it should be used 
> as sender. Otherwise existing logic of using the logged in user or 
> destination email address as the sender email address will remain. In 
> today's scenario this means all these emails will be treated as spam 
> as they will fail validation since the textpattern server may not have 
> the authorization to send email on behalf of other domains. 

If there is a need to force the From header to be identical to the 
envelope sender in some setups, why not add an extra config option to do 
so? I can see how that could be useful to satisfy DKIM/DMARC checks.

--Ruud

AmritasyaPutra:
I agree with you, they may be different as per protocol specification. It is still allowed, and valid today. At the same time, these emails will be marked as spam by most email servers. Also it is not logical to do so, it is confusing if not counter-intuitive. Why should I get an email 'from' myself when the comment is by another person? I would expect it to come from the website/article or the other person. DMARC validation is one more factor.

AmritasyaPutra:
Any further thoughts?

vanmelick:
On 03/07/2019 21:40, Jyoti Sharma wrote:
>
> I agree with you, they may be different as per protocol specification. 
> It is still allowed, and valid today. At the same time, these emails 
> will be marked as spam by most email servers. Also it is not logical 
> to do so, it is confusing if not counter-intuitive.
>

It is confusing to let the smtp_from preference have an additional 
meaning. Shouldn't there be an additional preference here "force From: 
address" or something similar, where you can specify an email address? 
That way it can be set separate from the envelope sender (which, IIRC, 
can't always be set).

> Why should I get an email 'from' myself when the comment is by another 
> person? I would expect it to come from the website/article or the 
> other person. DMARC validation is one more factor.
>

I agree with you here. At the very least the name should default to the 
sitename.

--Ruud

AmritasyaPutra:
Here is my analysis. The following are the fields set today:

Return-Path: admin@ekvastra.in
envelope-from: admin@ekvastra.in
To: author@gmail.com
From: Author Name <author@gmail.com>
Reply-to: commentwriter@gmail.com

Because ekvastra.in is not authorized to send emails "From" gmail.com all such emails will be marked as spam. Logically it should be commentwriter@gmail.com but this is also prohibited. And the comment writer just wrote a comment, not send an email so a notification email is a reasonable solution. The only sensible option here is to use the envelope-from as the from address also. This prevents the email from categorized as spam and also seems logical. 

I think users won't even get the difference between "SMTP envelope sender address" and "From". For example, the variable for the "SMTP envelope sender address" in code today is smtp_from. We can take a clue from what other cms do in such cases -- I think none offer more flexibility than "Send me an email update for new comments".

bloatware:
No further thoughts? Then I would suggest to merge it, since it seems to fix the related issue.

Bloke:
This is unfortunately above my pay grade. I'd defer to @vanmelick , since he is _way_ more knowledgeable in this field than I am.

Issue #1690, opened 2021-05-30
Textpattern 4.8.7 release flight plan


petecooper:
We are targeting Textpattern 4.8.7 for release in May 2021 to address some snags, a couple of regressions, and provide further refinements.

* [4.8.7 branch](https://github.com/textpattern/textpattern/tree/4.8.7)

Please update this post and subsequent replies with tasks etc as needed, assigning yourself or others where appropriate.

**Repo Issues**

- [x] textpattern/textpattern 
  * 4.8.7 milestone: [open issues](https://github.com/textpattern/textpattern/milestone/20) & [closed issues](https://github.com/textpattern/textpattern/milestone/20?closed=1)

## Personnel

* Release pilot: @petecooper 
* Release co-pilot: @Bloke 

## Preflight

- [ ] preflight
  - [x] latest translations (@philwareham)
  - [x] latest pophelp (@philwareham)
  - [x] update draft .com blog post [ID 440] to reflect combined changes of 4.8.6 and 4.8.7 (release pilot, or assigned person)
  - [x] write change history (everyone)
  - [x] tag public/admin themes with release tag (@philwareham)
  - [ ] tag Textpacks and Pophelps with release tag (@philwareham)

## Take Off

- [ ] take off
  - [x] package up for GitHub and .com (release pilot)
  - [x] update `latest.zip` and `latest.tar.gz` in server config (@petecooper)
  - [ ] update language pack links in server config (@petecooper)
  - [x] update demo site (@petecooper)
  - [x] publish .com blog post (release pilot, or assigned person)
  - [x] forum announcement (release pilot, or assigned person)
  - [x] forum feedback (release pilot, or assigned person)

## Inflight service

- [x] inflight service
  - [x] close milestone (release pilot, or assigned person)
  - [x] update `TXPversion` on `rpc.textpattern.com` (release pilot, or assigned person)
  - [x] update https://en.wikipedia.org/wiki/Template:Textpattern_version (@petecooper or @philwareham)
  - [x] socials (@petecooper or @philwareham)
  - [x] repin https://github.com/textpattern/textpattern/issues/968

petecooper:
Release post for 4.8.6 is revised (please edit article 440 without prejudice) to encompass 4.8.7 stuff.

I will start the release procedure for 4.8.7 in the next 30 minutes, please consider this a "down tools" request.

petecooper:
[...]

PR #1691, opened 2021-05-30
`release-4.8.7` to `main`

petecooper:
Should address https://github.com/textpattern/textpattern/issues/1690#issuecomment-851048576


Bloke:
Right, just had a chance to skim this. Looks like we can't merge it as-is, sadly, because it has the two extra commits after c9ad08a6. We can't have `main` going 'back to dev' as that'll scupper future merges (or at least the next one).

And on second thoughts I'm not sure about the change to the manifest version numbers for the admin-themes either, as it might cause merge problems next time. They might be fine, but I'm tempted to leave those out - they're largely inconsequential anyway.

No biggie. I'll manually cherry-pick all the commits up to and including c9ad08a6 into `main`, then close this PR.

petecooper:
Rats. Sorry @Bloke.

Bloke:
No worries. There's only a handful of commits to add.

Just trying to figure out why the 4.8.7 tag is in the wrong place (pointing to 4.8.6) and whether I need to move it by hand - and if so how to safely move it without making future dev commits go wonky. I would expect the tags to point to `main`, but if you browse them on GitHub and click the commit reference they go to the `dev` branch. Not what I expected, but maybe that's how git manages things internally and I don't need to worry about it - just move it on main and it' figures everything out on our behalf.

Issue #1766, opened 2021-12-25
Default pages & forms for new themes: wording tweaks


petecooper:
New theme assets have weird phrasing, e.g. default article form:

```
<!-- Default contents of the article tag goes here. See https://docs.textpattern.com/tags/article. -->
```

The default contents of the article _tag_ have nothing really to do with an empty artlcle _form_, since an empty article form won't show anything…but the defaults of the article tag when used correctly inside an article form _will_ show something.

Suggest we change this (and the other forms / pages) to:

```
<!-- Contents of the [form name] [form type] form goes here. -->
```

…with a link to the theme creation doc instead of the tag.

petecooper:
I can take this, I'm just not sure the best repo / place to do it in.

petecooper:
Done (scrappily).

PR #1769, opened 2021-12-29
tidy up & fix empty theme commentary

petecooper:
first pass at https://github.com/textpattern/textpattern/issues/1766

- Tidy up
- Remove erroneous references
- Remove links to specific tags in favour of docs landing page

Requesting review from @Bloke or @bloatware since it's PHP territory and the array escaping might not be best practice.

### DIFF
--- a/textpattern/vendors/Textpattern/Skin/Css.php
+++ b/textpattern/vendors/Textpattern/Skin/Css.php
@@ -41,7 +41,7 @@ class Css extends AssetBase implements CssInterface
     protected static $essential = array(
         array(
             'name' => 'default',
-            'css' => '/* Contents of the css tag goes here. See https://docs.textpattern.com/tags/css */'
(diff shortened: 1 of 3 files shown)


Bloke:
All looks good to me, thank you. Are we targeting this for the `dev` branch then? Since it's such a minor change it could be merged into 4.8.8 first, then get merged into dev as part of the regular sync process.

petecooper:
4.8.8 probably best - I can do manual sync up if that's easier, I wasn't sure how PRs are applied to branches.

Bloke:
I think you can target a PR to a particular branch in the same way you can against `dev` - just pick the branch first. Never tried it though. Doing it that way - going "up" from a release-ish branch to dev - just means less chance of conflicts when the release branches are periodically merged to dev.

petecooper:
All sorted, wanted to be sure I didn't tonk it in the process so it took a few more mins.

Issue #1786, opened 2022-02-08
4.8.8 / PHP 8.1: Deprecated “passing null to version_compare()” notice on Diagnostics panel


jools-r:
In **debug mode** on Textpattern **v4.8.8** and **PHP 8.1**, I get the following notice on the Diagnostics panel:

```
8192 "version_compare(): Passing null to parameter #2 ($version2) of type string is deprecated"
in/…/textpattern/include/txp_diag.php at line 849.
```
and:
```
adminErrorHandler()
textpattern/include/txp_diag.php:849 version_compare()
textpattern/include/txp_diag.php:213 checkUpdates()
textpattern/include/txp_diag.php:93 doDiagnostics()
textpattern/index.php:230 include()
```

[Line 849](https://github.com/textpattern/textpattern/blob/b16653990e5fac4bbe0397671709d0836bf9af1a/textpattern/include/txp_diag.php#L849) of txp:diag.php is the check against the pre-release version number. And indeed [version.json](https://textpattern.com/version.json) currently returns the following as we don't have a new pre-release at the moment:

```
{
  "textpattern-version": {
    "release": "4.8.8",
    "prerelease": null
  }
}
```

Presumably the version_compare statement just needs wrapping in a defensive if $prerelease is not null wrapper...

Bloke:
Yes, or we patch the json value to return 'false' or something instead, which means no changes to core required :)

petecooper:
Should I be worried about `version.json` going from null to Boolean?

bloatware:
I guess it would complain about anything else than a string.

Bloke:
So can we return `''` then, to mean "not available"?

bloatware:
Dunno, just saying php 8.1 seems to expect the exact function [signature](https://www.php.net/manual/en/function.version-compare.php). Which implies hundreds of potential warnings to patch.

PR #1794, opened 2022-03-28
Update txp_diag.php

jools-r:
Silence PHP 8.1 – Deprecated “passing null to version_compare()” notice on Diagnostics panel when no prerelease exists.

- only perform prerelease comparison when $prerelease is not null

Addresses issue #1786.

### DIFF
--- a/textpattern/include/txp_diag.php
+++ b/textpattern/include/txp_diag.php
@@ -846,7 +846,7 @@ function checkUpdates()
                 $lastCheck['msgval'] = array('{version}' => $release);
             }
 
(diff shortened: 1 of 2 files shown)


philwareham:
@Bloke @bloatware is this patch fine to be merged?

jools-r:
I've updated the PR accordingly. Thanks for the feedback Oleg.

Bloke:
Looks good but I'd rather not merge this to main, since that's for releases. Could you please raise this PR against dev? Thanks.

Issue #1798, opened 2022-04-20
CLS issue points at image element not having width and height


AmritasyaPutra:
I am referring to this location:
https://github.com/textpattern/textpattern/blob/d4f026d7f675d148ab1896da3e35190ca34be5f0/textpattern/setup/themes/four-point-nine/forms/misc/images.txp#L9

Existing code in the above location renders the following on my website:
```
 <p itemprop="image" itemscope="" itemtype="https://schema.org/ImageObject">
        <img loading="lazy" itemprop="url contentUrl" src="https://ekvastra.in/images/29.png" alt="Anger" >
        <meta itemprop="width" content="700">
        <meta itemprop="height" content="700">
 </p>
```
 
For whatever reason, pagespeed Insights, GTmetrix, and dareboost all were saying that "Image elements do not have explicit width and height " ... "Set an explicit width and height on image elements to reduce layout shifts and improve CLS". I checked in Firefox that the itemprop width and height were greyed out.

I specified the width and height in the img element instead of separate meta line and the problem was fixed! 

I also removed lazy loading because I was getting "Largest Contentful Paint image was lazily loaded" ... "Above-the-fold images that are lazily loaded render later in the page lifecycle, which can delay the largest contentful paint." This isn't important here though.

cara-tm:
FYI,, to keep the "lazy-loading" for images below the fold but to remove this attribute above the fold, I did this model within a website:


    
    <txp:variable name="img_counting" add />
    
    <txp:variable name="caption" value='<txp:image_info />' />
    
    [...]
    
    <figure itemprop="image" itemscope itemtype="https://schema.org/ImageObject">
    
    	<img itemprop="url contentUrl"<txp:evaluate query='number(<txp:variable name="img_counting" />) < 2'><txp:else /> decoding="async" loading="lazy"</txp:evaluate> src="<txp:image_url link='0' />" alt="<txp:image_info type='alt' />" width="<txp:image_info type="w" />" height="<txp:image_info type="h" />">
    
    [...]

</figure>

jools-r:
> For whatever reason, pagespeed Insights, GTmetrix, and dareboost all were saying that "Image elements do not have explicit width and height " ... "Set an explicit width and height on image elements to reduce layout shifts and improve CLS". I checked in Firefox that the itemprop width and height were greyed out. I specified the width and height in the img element instead of separate meta line and the problem was fixed!

[...]

PR #1799, opened 2022-04-20
Possible fix for CLS due to image element not having width and height

AmritasyaPutra:
This is in context of this issue: https://github.com/textpattern/textpattern/issues/1798

The details are present in the reported issue.

Specify the width and height in the img element instead of separate meta line.

Remove lazy loading because of "Largest Contentful Paint image was lazily loaded" ... "Above-the-fold images that are lazily loaded render later in the page lifecycle, which can delay the largest contentful paint."

For below-the-fold images the browser will do the magic and take care that it does not affect the above-the-fold screen getting ready on time. We can forego the lazy loading tag here as it affects web vitals.

### DIFF
--- a/textpattern/setup/themes/four-point-nine/forms/misc/images.txp
+++ b/textpattern/setup/themes/four-point-nine/forms/misc/images.txp
@@ -5,18 +5,14 @@
 <txp:if_variable name="caption" value="">
 
     <p itemprop="image" itemscope itemtype="https://schema.org/ImageObject">
-        <img loading="lazy" itemprop="url contentUrl" src="<txp:image_url link='0' />" alt="<txp:image_info type='alt' />">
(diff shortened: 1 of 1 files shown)


philwareham:
The image form is a kind of 'catch-all' bit of code that can cater to general images - not necessarily the best solution for all use cases - hence why it doesn't enforce widths/heights.

Agree that the lazy landing is now superfluous - all the main browsers now support auto lazy loading using their own analysis of the page, so there is not much benefit from enforcing loading with an attribute.

TLDR; I will amend this for the next release, thanks. BTW the theme for Textopattern has its own repo which can be found here: https://github.com/textpattern/textpattern-default-theme

AmritasyaPutra:
Okay, thanks for the update, I get your point.

PR #1804, opened 2022-04-27
Update txp_diag.php / Silence passing null notice

jools-r:
Silence PHP 8.1 – Deprecated “passing null to version_compare()” notice on Diagnostics panel when no prerelease exists.

- only perform prerelease version comparison when $prerelease is set and not null.

Addresses issue #1786.


jools-r:
If you can correct my stupid ntoice->notice typo in the commit text when accepting the PR, please do. Thanks!

PR #1809, opened 2022-05-13
Implemented metaWeblog.newMediaObject

Melonking906:
Allows for image upload over the MovableType API, thanks to @Bloke @redsweater @jakob for the help!


Bloke:
Looking good, thanks @Melonking906. We can give this a thorough grilling with various file types (including our fledgling support for formats like AVIF and WEBP) but it seems a pretty solid implementation from first comb. Nice one.

PR #1817, opened 2022-07-25
Rotate/flip thumb if src img has exif orientation

jools-r:

- Add check for EXIF orientation metadata in source image
- If yes, rotate thumbnail accordingly + swap height / width values for further processing
- If yes, flip thumbnail accordingly if necessary

### DIFF
--- a/textpattern/lib/class.thumb.php
+++ b/textpattern/lib/class.thumb.php
@@ -265,32 +265,34 @@ class wet_thumb
             $this->_SRC['width'] = 100;
         }
 
-        // Check exif orientation of source image.
-        $exif = exif_read_data($this->_SRC['file']);
-        if (!empty($exif['Orientation'])) {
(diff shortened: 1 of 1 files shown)


bloatware:
Thank you, @jools-r . I'm absolutely fine with merging it, but can only push the green button, having no recent camera to test. @Bloke?

jools-r:
Actually, I did notice an error message in some browsers where a file format does not handle exif orientation data, e.g. GIF or PNG. Currently only JPEG and TIFF support EXIF orientation data, and of those two we only accept JPEG as an image source. WebP and AVIF purportedly have their own implementations – see [zpl.fi/exif-orientation-in-different-formats/](https://zpl.fi/exif-orientation-in-different-formats/).

I'll update the patch so that it only acts on incoming files in JPEG format.

jools-r:
Thanks Pete!

PR #1844, opened 2022-11-10
Self-closing tag ending acc. to doctype setting

jools-r:

Appease the W3C validator…

txp_tag.php
- class BuilderTags (tag builder for html image code)

txplib_html.php
- function: tag_void (used in fInput)
- function: doWrap for br/hr break attributes
- function: doTag when content is empty
- function: doLabel break

comment.php
- function: input (for comments)

taghandlers.php
- function: txp:image
- function: txp:article_image
- function: txp:meta_keywords
- function: txp:meta_description
- function: txp:meta_author
- function: txp:feed_link / link_feed_link
- searchinput + popup
- rsd link

(this patch deviates only slightly from the corresponding patch for the main branch as function doWrap has changed slightly in the dev branch)

### DIFF
--- a/textpattern/include/txp_tag.php
+++ b/textpattern/include/txp_tag.php
@@ -2351,7 +2351,7 @@ class BuilderTags
 
                     $out .= $this->tdb(
                         ($wraptag ? "<$wraptag>" : '').
-                        '<img src="'.$url.'" width="'.$w.'" height="'.$h.'"'.$alternate.$cap.$htmlid.$cls.$inlinestyle.(get_pref('doctype') == 'html5' ? '>' : ' />').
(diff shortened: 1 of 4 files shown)


Bloke:
Stellar work, Sir. Thank you.

Would you mind just whipping through for a bit of house style tweaking please and standardise on use of single quotes for 'doctype' rather than double quotes in a few places?

Also, and this is probably not a concern given we use it all over the place, I'm not sure if we should be using `===` for strict comparison here instead of `==`. Essentially it makes no difference and saves some bytes to use the double equals. The purist in me says we ought to be using strict comparison wherever we can as we gradually improve the codebase. But it depends if the output of `get_pref()` always returns a string. If there's a chance it doesn't (e.g. it might return false or null), then for the time being let's stick with `==` otherwise it'll introduce hard to track down bugs.

Note to self: cast the output of library functions like `get_pref()` if not already.

jools-r:
The inconsistent quotes are corrected and PR is updated. Just read that I overshot your requirements and included type comparison, but there are similar instances with `===` already in the code.

Bloke (approved):

PR #1848, opened 2022-11-20
Add support for images in SVG format

osadl:
SVG support could easily be added by providing some short wrappers to the related GD library functions.


bloatware (changes requested):
Thank you. I wouldn't check for `<script />` tag here, because it does not protect against various `on*` events xss, and an admin must know what he is doing anyway. But it would be nice to extract stand-alone `<svg />` images from `$content`, which otherwise could be a general type xml.

osadl:
Created various versions of encapsulated SVGs for testing such as

<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<html>
<head>
</head>
<body>
<svg>
...
</svg>
</body>
</html>

Both full image and thumb had in all test cases the encapsulation removed, and the only remaining tree element was <svg>. Is this what you wanted me to implement?

osadl:
My test cases were for example
```
<html>
<head>
</head>
<body>
<svg>
  [..]
</svg>
</body>
</html>
```

bloatware:
Yes, thank you, this will be more robust when displaying svg images via `` tag.

Awaiting @Bloke opinion, but all this makes me think that acceptable image formats are too tightly integrated in core. There should be a way to pluginify it.

bloatware:
So, ready for the prime time? @Bloke?

Bloke:
If everyone else is happy, I'm fine with it. I haven't tested it owing to a lack of SVG files knocking around.

I say merge it and we'll deal with anything that crops up in wider testing. Many thanks, @osadl

Issue #1933, opened 2024-07-18
A txp:tag to create a page-wide CSP nonce


jools-r:
On the admin-side we have the constant TEXTPATTERN_CSP_NONCE and the `\Textpattern\UI\Script` and `\Textpattern\UI\Style` methods to output `script src`, `script`, `style` and `link rel="stylesheet"` tags with a corresponding nonce. Having just made updates to bot_wtc and glz_custom_fields, it works well.

On the forum and elsewhere here in the issue, we've discussed different ways of tapping into apache or nginx generated unique_id's for using as a nonce in the public side, but as Stef has shown, it works just as well if generated by txp and then used throughout a page template. Up to now, I've set this as a variable at the top of a page and then construct the style and script tags accordingly. For my own templates, that's fine, but when making a template, you start to build in conditionals for if it is available.

Is there mileage for have a `txp:nonce` tag (or whatever) for setting a csp nonce that other tags (e.g. txp:css) could 'contextually recognise' and adapt their output accordingly?

I wonder if needs some kind of “contextual trigger”, for it to activate. In non-CSP use, it's not needed, only when csp headers have been set is it relevant. On the admin side you set that via config.php, but presumably front-end and admin-side CSP could potentially differ, so that activating flag can't be used for both.

Bloke:
I like this idea. And it's an absolute cinch to implement in core. You might like to investigate the world's smallest plugin, smd_token:

```
# smd_token v0.1.0
# Generate cryptographically random tokens for use within Textpattern
# Stef Dawson
# https://stefdawson.com/

# ......................................................................
# This is a plugin for Textpattern - http://textpattern.com/
# To install: textpattern > admin > plugins
# Paste the following text into the 'Install plugin' box:
# ......................................................................

H4sIAAAAAAAAA+1WS2/jNhC+61cM1ACyA1u2nDiOlTpIsEEfQBcNNtn2EsCgpZFEhKa0JFXb
WOS/d0hJjl1sFkUvvUQXkaPhzDffPCgWR5P4q47PY1+yNfpXOp7Hvl6nS1M+o7T7Wez/hUrz
0u2msT8Oo3Bs1xexz2pTlMpuoij2HwxmcMc2utGNxp3CslbcSiZnsV8YU+l4NNKknDrdMCnX
o85GijpRvDKtvwty/zNKVMwgJGpXmTJXrCp4woTYgWIyLdfgwGrISgW1RthwU3AJj7g1FTMG
[...]

Issue #1394, opened 2019-03-09
New properties for the native css tag


cara-tm:
### Is your feature request related to a problem?

No (an improvement).

### What is the feature?

To prevent attacks, possibly into CSS files (see: https://www.mike-gualtieri.com/posts/stealing-data-with-css-attack-and-defense), maybe it's time to add new properties for the `<txp:css />` tag:
"`integrity`" attempted to store a SRI hash (Subresource Integrity) and its compagnon "`crossorigin`".

Sample of a possible use:

`<txp:css name="print" format="flat.link" media="print" integrity="sha384-/VLZ9QhATgiUs4ha2DcLiKyQ40fBFB+Wub8YYfGE4GQWPTB1glePJzCoX9Iqzi0q" crossorigin="anonymous" />`

Online tools exist to get that hash (https://www.srihash.org/), maybe a more simple way could be added into core to generate on the fly a such code if user set the property on "1":

`<txp:css name="print" format="flat.link" media="print" integrity="1" crossorigin="anonymous" />`

bloatware:
Something is [planned](https://github.com/textpattern/textpattern/tree/integrity) in this sense, for JS too. But for flat links this requires some server configuration, since PHP is totally bypassed. Needs investigation.

cara-tm:
Fine. Cool ;)

Bloke:
I like the `format` idea. Adds serious flexibility.
Regarding CSP, see https://github.com/textpattern/textpattern/issues/1933#issuecomment-2238495192

PR #1938, opened 2024-08-11
txp:css format attributes for inline output

jools-r:

- format="" attribute for direct css output.
- format="inline" attribute for output in a <style> tag.
- support for escape attributes for inline output.

The "inline" option is self-explanatory. The empty option allows one to construct an own style tag, for example with a nonce, to construct a style block out of several stylesheets, or to output a stylesheet twice, as is common, for example, when providing dark mode variables for system-set dark mode (e.g. via @media) or user-set dark mode (e.g. via class or data-attribute).

The query is saved in a static variable should the same stylesheet be output more than once.

## Background

The underlying motivation is to allow stylesheet entries in _Presentation › Styles_ to be used as user-editable theme overrides containing CSS custom properties (CSS variables). Often only a few variables are overridden, e.g. brand colours, font-family etc., and are then more easily inserted inline directly in the page head.

See also issues #1394 and #1933 for further context.

Improvements welcome!

### DIFF
--- a/textpattern/publish/taghandlers.php
+++ b/textpattern/publish/taghandlers.php
@@ -240,7 +240,7 @@ function page_title($atts)
 function css($atts)
 {
     global $css, $doctype, $pretext;
-    static $stylebody = null;
+    static $stylebody = array();
 
     extract(lAtts(array(
         'escape' => true,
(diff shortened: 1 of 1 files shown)


Bloke:
Thank you for providing this. Looks like a sensible addition. A couple of comments:

1. I'm not sure how finicky PHP is (or is going to become) about having a static variable defined null and then redefined as an array later. Is it better to define it as `static $stylebody = array();`? I think PHP allows empty initialisation like that. Pretty sure it can't be done for globals, but I think it's okay for statics.
2. It might make sense to run `$name` and `$theme` through one of our `sanitiseFor...()` routines in the conditional block, to guard against the prospect of either value containing invalid chars that would break the array index. I think we have a function that is safe for variable usage but I can't remember offhand. If we don't have such a routine, we could do with one (I think there's one in the Skin class that we could utilise if not). I'm fairly certain PHP variables still can't begin with numbers or contain dashes (among other symbols) so if you had a stylesheet called `1st-sheet` it would break the code. Give it a whirl and see what happens!

Thank you so much for this. It'll be a really handy addition to squeeze into 4.9.0.

jools-r:
Thanks for the tips here and via email. Both 1. and 2. are incorporated. 2 uses the same function used to sanitize theme, page, form and style names in the Skin class as you suggested.

FWIW: I didn't experience any problems with style names like @1st-sheet@, @1st#sheät@ both with and without sanitization, whether set in the style panel and referenced automatically or supplied via the tag attribute. The $debug flag shows that the characters come through in the query. My db is utf8mb4, so perhaps the sanitization catches problems with other encodings. Adding sanitization does correct style attribute values with a space in the name, though the styles panel also sanitizes that when naming a style.

petecooper:
Polite bump on this - is it in scope for 4.9.0, sirs?

PR #1940, opened 2024-08-15
Strip some phpmailer ancillary files to save a few bytes

petecooper:

- removed non-essential stuff from phpmailer to save a few bytes

It's not clear if https://github.com/textpattern/textpattern/blob/dev/textpattern/vendors/phpmailer/phpmailer/composer.json is essential for the production release. If it can go, it could be zapped here or removed as part of the toolbelt release process.

Adding @Bloke for sanity check (not urgent).


petecooper:
Related dumb questions -  do we need the .js _and_ the associated .map in https://github.com/textpattern/textpattern/tree/dev/textpattern/vendors/cure53/DOMPurify/dist ? I can see the logic with dev / debugging, but I don't know the importance / relevance to the production release.

Bloke:
This is all good stuff. If there's anything like these files that can be stripped out of any third party libraries, they can be added to the txp-gitdist.sh script. There's a whole raft of deletions done there as part of the bundling process.

PR #1971, opened 2025-02-05
RPC Improvements

Melonking906:

- Fixes for RPC under PHP 8+
- More robust image upload support via RPC
- Fixes to Body_HTML data not being populated by RPC edits
- Proper support for post status in RPC clients
- Fixes to timezone error in RPC
- Fixes to error in the event a category is not set on a post in RPC
- Support for auto settings detection for RPC clients added to default template

### DIFF
--- a/rpc/TXP_RPCServer.php
+++ b/rpc/TXP_RPCServer.php
@@ -36,6 +36,11 @@ require_once txpath . '/lib/txplib_misc.php';
 
 define('IMPATH', $path_to_site . DS . $img_dir . DS);
 
+// HTTP address of the site serving images.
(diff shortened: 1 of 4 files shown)


Bloke:
This is all brilliant stuff, thank you. Glad you spotted this.

One minor thing: please would you tweak the PR to avoid using the null coalescing `??` operator for now, since our [minimum ride height is PHP 5.6.0](https://github.com/textpattern/textpattern/blob/dev/textpattern/lib/constants.php#L738). I know it's unlikely anyone will be running that, but that's our minimum target for this release.

Many thanks.

PR #2007, opened 2025-08-15
Add $from attribute to txpMail()

jools-r:

- Ability to supply a user-defined sender $from address as a string, or
- Ability to supply a user-defined sender email address and name as $from array
- Existing behaviour remains if not supplied

## Reasoning

The current behaviour is:

* If a txp user is logged in, send emails in their name but using the `publisher_email` preferences. If not available, use the user's email.
* If the txp user is not logged in, use the receiver's email and look up if the name is already available in the registered admin users.

Now that we are able to supply SMTP credentials and mail out using PHPMailer, this can cause conflicts with providers that limit the sending of emails to that email address for spam prevention reasons. Mails sent from the publisher_email or from the recipient won't be sent (smtp_log shows that my provider rejects the sender email).

This patch allows you to pass a from email to txpMail().

## Requests / notes

* $RealName is not currently escaped to avoid people with apostrophes in names being mangled. I tried passing a javascript alert via that field and had no problems on the receiving end in my email program or webmail program. We do little escaping of RealName elsewhere. Okay?
* There is a potentially uncaught situation where the from address has an invalid syntax and the publisher_email is likewise invalid/incomplete. Down to the user to get it right?

### DIFF
--- a/textpattern/lib/txplib_misc.php
+++ b/textpattern/lib/txplib_misc.php
@@ -2351,38 +2351,56 @@ function is_valid_email($address)
  * }
  * </code>
  *
- * @param   string $to_address The receiver
- * @param   string $subject    The subject
- * @param   string $body       The message
- * @param   string $reply_to   The reply to address
+ * @param   string $to_address  The receiver
+ * @param   string $subject     The subject
+ * @param   string $body        The message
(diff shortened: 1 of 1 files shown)


jools-r:
Thanks 👍

PR #2040, opened 2025-12-25
Always send system mails from publisher_email

jools-r:

If no $from attribute is specified, send system emails using txpMail() from:
* Sender name = ~~logged in user or, when no user is logged in, the recipient~~ site name
* Sender email = publisher_email pref

If the publisher_email is invalid or unset, use no-reply@domain.com


Bloke:
@jools-r could you do us a huge favour though please and submit this against the v4.9.1 branch? We can then suck the changes into dev next time it's synced. Many thanks.

